Issue #1, opened 2022-01-10
Configure build process


kirillt:
Depends on https://github.com/ARK-Builders/ARK-Navigator/issues/163
The library must be built in 2 modes: _Debug_ and _Release_, both must be published and downloadable from GitHub Actions.
ARK Navigator and other dependents should be configured to use _Release_ build by default.

sisco0:
- [ ] Create the project structure and fill it with the current implementation
- [ ] Fill the `README.md` file with development and deployment instructions
- [ ] Create a CI system under _Github Actions_

kirillt:
Implementation of this task required commits in different repos:

* [The Library](https://github.com/ARK-Builders/arklib):
[90ca27ace9000d943a18bbcedac32b56d23c3a99](https://github.com/ARK-Builders/arklib/commit/90ca27ace9000d943a18bbcedac32b56d23c3a99)
[052832dd56d419b742033bd8e5bf7ea8ceae9944](https://github.com/ARK-Builders/arklib/commit/052832dd56d419b742033bd8e5bf7ea8ceae9944)

* [Android wrapper](https://github.com/ARK-Builders/arklib-android/):
[56865af62f112668039de922b1f630f9299e948e](https://github.com/ARK-Builders/arklib-android/commit/56865af62f112668039de922b1f630f9299e948e)

* [ARK Navigator](https://github.com/ARK-Builders/ARK-Navigator):
[8c2288ab8955e97ee14783145b617cf19474ad16](https://github.com/ARK-Builders/ARK-Navigator/commit/8c2288ab8955e97ee14783145b617cf19474ad16)

This repo can be used as pure Rust dependency in other apps or libs. It can be used in pure Rust apps.
First, CLI tool and lightweight Web viewer will be implemented using it.
After, we might reach Linux/iOS/Windows platforms.

The lib is used in [Android wrapper](https://github.com/ARK-Builders/arklib-android/) and is wrapped into AAR package which can be embedded into any Android app without necessity to install Rust and NDK. We need to publish it to a repo like Maven Central for easier versions management (right now, manual placement into Gradle project is necessary).

PR #4, opened 2022-01-12
Project basic setup

sisco0:
This first PR bases itself on creating the project structure.
The following list of features have been created:

- [x] Main Rust library
- [x] Android bindings
- [x] Naive test for compute_id main Rust library method

References #1


kirillt:
We also need simultaneous PR in ARK-Navigator which removes Rust code from there and imports `arklib`.

Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+crate-type = ["dylib"]
+
+[dependencies]
+log = "0.4.14"
+crc32fast = "1.3.0"
+env_logger = "0.9.0"
+
+[target.'cfg(target_os="android")'.dependencies]
+jni = { version = "0.19.0", default-features = false }
+android_logger = "0.10.1"
```
kirillt, line 17: Can we leave Android-related stuff in ARK-Navigator repo?

src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+        trace!("{} bytes has been read", bytes_read);
+        trace!("checksum: {:#02x}", checksum);
+        assert!(bytes_read == file_size.try_into().unwrap());
+        return checksum;
+    }
+}
+
+#[cfg(target_os="android")]
+#[allow(non_snake_case)]
+pub mod android {
```
kirillt, line 45: Also this

src/lib.rs
```diff
@@ -0,0 +1,94 @@
+pub mod resource_id {
+    use std::fs;
+    use std::path::Path;
+    use std::io::{BufReader, BufRead};
+    use crc32fast::Hasher;
+    use log::trace;
+    pub fn compute_id<P: AsRef<Path>>(
+        file_size: usize,
+        file_path: P) -> i64 {
```
kirillt, line 9: If we need `i64` only for passing into Kotlin, I would create some adapter on that side.
We should use `u32` here if possible.

kirillt:
Will be followed-up by additional commits in order to close #1

Issue #15, opened 2022-06-16
Persisted/Replicated index


kirillt:
See https://github.com/ARK-Builders/ARK-Navigator/issues/142 for the context.
**TL;DR**: At the moment, index is built for any "root" folder and stored in memory.

Would be cool to persist it, so we wouldn't recalculate resource ids on different devices. If we store the index into our `.ark` folder, then it cache gets synced to other devices (that's why _replicated_). Smart write to avoid conflicts is necessary, but would be too difficult for this moment, let's assume now that only 1 device writes index into some root at the same moment.

Let's fix `.ark/index` path for it.

kirillt:
This is the only part of the storage which is not user-defined and allowed to be lost.

kirillt:
Reminder: `.ark` folder is our "persisted and replicated storage" which gets synced to all other devices using [Syncthing](https://syncthing.net/). Later, we might implement our own sync mechanism.

PR #23, opened 2022-08-27
#15: Android Index Rust

nagato-chan:
#15

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -37,12 +37,27 @@ impl ResourceIndex {
         root_path: P,
         resources: HashMap<CanonicalPathBuf, ResourceMeta>,
     ) -> Self {
-        Self {
-            path2meta: resources,
+        // TODO Return Result
+        log::info!("creating resource index from giving resources");
+        let root = CanonicalPathBuf::canonicalize(root_path).unwrap();
+
+        let mut index = Self {
+            path2meta: HashMap::new(),
             collisions: HashMap::new(),
             ids: HashSet::new(),
-            root: root_path.as_ref().to_path_buf(),
+            root: root.into_path_buf(),
+        };
+        for (path, meta) in resources {
+            add_meta(
+                path,
+                meta,
+                &mut index.path2meta,
+                &mut index.collisions,
+                &mut index.ids,
+            );
         }
+        log::info!("Index created from giving resources.");
+        index
     }
     pub fn calc_diff(&self) -> Difference {
         let (present, absend): (Vec<_>, Vec<_>) = self
--- a/src/index/cache.rs
+++ b/src/index/cache.rs
@@ -1,0 +1,28 @@
+use crate::id::ResourceId;
+use canonical_path::CanonicalPathBuf;
+use std::str::FromStr;
+
+#[derive(Debug)]
+pub struct IndexCache {
+    pub path: CanonicalPathBuf,
+    pub id: ResourceId,
+}
+
+impl ToString for IndexCache {
+    fn to_string(&self) -> String {
+        todo!()
+    }
+}
+
+// impl FromStr for IndexCache {
+//     type Err = ;
+//     fn from_str(s: &str) -> Result<Self, Self::Err> {
+//         todo!()
+//     }
+// }
+
+impl IndexCache {
+    pub fn parse(input: &str) -> Result<&str, Self> {
+        todo!()
+    }
+}
--- a/src/index/mod.rs
+++ b/src/index/mod.rs
@@ -1,5 +1,8 @@
 pub mod cache;
-use std::path::{Path, PathBuf};
+use std::{
+    fs,
+    path::{Path, PathBuf},
+};
 
 use std::time::SystemTime;
 use std::{
@@ -17,7 +20,7 @@ use nom::{sequence::preceded, IResult};
 use crate::id::ResourceId;
 use crate::meta::ResourceMeta;
 
-pub const INDEX_CACHE_DIR: &'static str = ".ark/index/cache";
+pub const INDEX_CACHE_DIR: &'static str = ".ark/index";
 
 pub const INDEX_DIR: &'static str = ".ark/index";
 
@@ -29,6 +32,7 @@ pub struct ResourceIndex {
     pub collisions: HashMap<ResourceId, usize>,
     // id records for hash collisions check
     ids: HashSet<ResourceId>,
+    // the root path
     root: PathBuf,
 }
 // Since we treat outdated index as deleted index and updated index as added index,
@@ -39,13 +43,14 @@ pub struct IndexUpdate {
     pub added: HashMap<CanonicalPathBuf, ResourceMeta>,
 }
 
-pub struct Difference {
-    pub updated: Vec<CanonicalPathBuf>,
-    pub deleted: Vec<CanonicalPathBuf>,
-    pub added: Vec<CanonicalPathBuf>,
-}
+// pub struct Difference {
+//     pub updated: Vec<CanonicalPathBuf>,
+//     pub deleted: Vec<CanonicalPathBuf>,
+//     pub added: Vec<CanonicalPathBuf>,
+// }
 
 impl ResourceIndex {
+    /// initial index from given path,
     pub fn from_resources<P: AsRef<Path>>(
         root_path: P,
         resources: HashMap<CanonicalPathBuf, ResourceMeta>,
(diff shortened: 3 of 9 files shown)


src/meta.rs
```diff
@@ -34,32 +38,150 @@ impl ResourceMeta {
...
     None
 }
+#[derive(Deserialize, Serialize)]
+pub struct LinkFile {
+    pub title: String,
+    pub desc: String,
+    pub url: String,
+}
+
+fn parse_link(path: &CanonicalPathBuf) -> Result<ResourceKind, Error> {
```
kirillt, line 174: We have Link-to-Web in `arklib` now, pull latest commits from `main`.

src/index.rs
```diff
@@ -62,28 +120,39 @@ impl ResourceIndex {
         log::trace!("Known paths:\n{:?}", self.path2meta.keys());
 
         let curr_entries = discover_paths(self.root.clone());
-
-        //assuming that collections manipulation is
+        // assuming that collections manipulation is
         // quicker than asking `path.exists()` for every path
         let curr_paths: Paths = curr_entries.keys().cloned().collect();
         let prev_paths: Paths = self.path2meta.keys().cloned().collect();
+        // Union set of curr_paths and prev_paths (path already existed, but may be updated)
```
kirillt, line 127: It is not union, it is intersection.

src/index.rs
```diff
@@ -20,11 +21,68 @@ pub struct ResourceIndex {
 
 #[derive(Debug)]
 pub struct IndexUpdate {
-    pub deleted: HashSet<ResourceId>,
+    pub updated: HashMap<CanonicalPathBuf, ResourceMeta>,
```
kirillt, line 24: An update is deletion followed by addition, because `ResourceId` changes, so technically new version is a new resource. Two fields `added` and `deleted` should be enough. This is different in ARK Navigator which should be considered older version.

src/index.rs
```diff
@@ -62,28 +120,42 @@ impl ResourceIndex {
...
+                .iter()
+                .filter_map(|(path, meta)| {
+                    if !preserved_paths.contains(path.as_canonical_path()) {
+                        Some((path, meta))
+                    } else {
+                        None
+                    }
+                })
+                // If the path was included in prev_path, it should had been deleted from device.
+                .partition(|(path, meta)| {
```
kirillt, line 144: Mm, not sure if this is correct.

src/index.rs
```diff
@@ -111,35 +183,48 @@ impl ResourceIndex {
...
+                if let Some(meta) =
+                    self.path2meta.remove(path.as_canonical_path())
+                {
+                    self.ids.remove(&meta.id);
                 } else {
                     log::warn!("Path {} was not known", path.display());
                 }
             });
 
+        // treating deleted and updated paths as deletions
```
kirillt, line 207: But why do you want to change this function at all?

src/meta.rs
```diff
@@ -34,32 +38,150 @@ impl ResourceMeta {
...
         Ok((path.clone(), meta))
     }
 }
 
-//todo
-pub type ResourceKind = ();
-pub type ResourceExtra = ();
+#[derive(Eq, PartialEq, Hash, Clone, Debug, EnumString, Display, EnumCount)]
+#[strum(ascii_case_insensitive)]
+pub enum ResourceKind {
```
kirillt, line 59: I think we should port resource kinds in the last turn... the same as previews. First, just plain interaction with Android should be set up.

kirillt: We can just pass `None` before we implement resource kinds.
```
val kind: ResourceKind?
```

src/meta.rs
nagato-chan: Got. Will merge it later.

src/meta.rs
nagato-chan: Agree.

src/index.rs
```diff
@@ -73,7 +131,8 @@ impl ResourceIndex {
             .collect();
 
         let created_paths: HashMap<CanonicalPathBuf, DirEntry> = curr_entries
-            .iter()
+            .clone()
+            .into_iter()
```
kirillt, line 135: _to check_

src/meta.rs
```diff
@@ -2,24 +2,24 @@ use crate::id::ResourceId;
...
 use std::ffi::{OsStr, OsString};
-use std::time::SystemTime;
+use strum::{Display, EnumCount, EnumString};
 use walkdir::DirEntry;
-
 #[derive(Eq, PartialEq, Hash, Clone, Debug)]
 pub struct ResourceMeta {
     pub id: ResourceId,
-    pub modified: SystemTime,
+    pub modified: DateTime<Utc>,
```
kirillt, line 13: Why `DateTime` is better than `SystemTime`?

src/index.rs
```diff
@@ -111,14 +170,16 @@ impl ResourceIndex {
...
                     }
                 }
             })
             .collect();
 
-        let mut deleted: HashSet<ResourceId> = HashSet::new();
+        let mut deleted = HashMap::new();
 
         // treating deleted and updated paths as deletions
         prev_paths
```
kirillt, line 185: _Maybe `indexed_path` would be a better name._

kirillt: _Observation:_
preserved_paths = prev_paths \/ curr_paths
prev_paths - preserved_paths = prev_path - curr_path _= deleted?_

_Renaming suggestions:_
prev_paths -> index_paths
curr_paths -> fs_paths
preserved_paths -> common_paths

src/index.rs
```diff
@@ -133,25 +194,25 @@ impl ResourceIndex {
...
                     log::info!("The same for new paths");
-                    scan_metadata(created_paths).into_iter()
+                    scan_metadata(&created_paths).into_iter()
                 })
                 .filter(|(_, meta)| !self.ids.contains(&meta.id))
                 .collect();
 
         for (path, meta) in added.iter() {
-            if deleted.contains(&meta.id) {
+            if deleted.contains_key(path) {
```
kirillt, line 215: No, this doesn't look correct. Log message bellow: "Resource _id_ was moved to _path_".

If `path` is contained in both `added` and `deleted` then we have a mistake in their calculations and must emit something like `AssertionError`.

Resource is indeed moved only if the a resource with `meta.id` was deleted at `path1` and was added at `path2`.

src/index.rs
```diff
@@ -466,13 +527,14 @@ mod tests {
 
             let mut file_path = path.clone();
             file_path.push(FILE_NAME_1);
-            std::fs::remove_file(file_path)
+            let updated_path =
```
kirillt, line 530: This doesn't make sense :) It's a unit test, all seems to be ok here.

src/meta.rs
```diff
@@ -34,28 +34,87 @@ impl ResourceMeta {
...
+        title: Option<String>,
+        description: Option<String>,
+        url: Option<String>,
+    },
+
+    PlainText,
+    Archive,
+}
+
+// Currently all unrecognized/unsupported format will be parsed to PlainText
```
kirillt, line 75: Unrecognized resource kind must be `None`.

src/meta.rs
```diff
@@ -34,28 +34,87 @@ impl ResourceMeta {
...
+impl From<CanonicalPathBuf> for ResourceKind {
+    fn from(path: CanonicalPathBuf) -> Self {
```
kirillt, line 77: _to check later_

kirillt (changes requested):

src/index/cache.rs
```diff
@@ -0,0 +1,28 @@
+use crate::id::ResourceId;
+use canonical_path::CanonicalPathBuf;
+use std::str::FromStr;
+
+#[derive(Debug)]
+pub struct IndexCache {
```
kirillt, line 6: I don't think this file is useful. You should just dump `ResourceIndex` from `index/mod.rs` into a file, field `collisions` can be ignored while writing to disk. Field `ids` can be ignored as well, only `path2meta` matters. Field `root: PathBuf` is what point to where `.ark` subfolder must be used.

Actually, other fields of `ResourceMeta` should be written as well (initially, I thought the opposite). So, not only `id` but also filesize, modified time etc.

Another direction: while reading `.ark/index` each line `<path>: <id>, <modified>, ...` must create an entry with values of types `CanonicalPathBuf` into `ResourceMeta` in the `ResourceIndex`.

src/index/mod.rs
```diff
@@ -1,27 +1,40 @@
...
 use walkdir::{DirEntry, WalkDir};
 
 use anyhow::Error;
 use log;
+use nom::{sequence::preceded, IResult};
 
 use crate::id::ResourceId;
 use crate::meta::ResourceMeta;
 
+pub const INDEX_CACHE_DIR: &'static str = ".ark/index/cache";
```
kirillt, line 20: It's not cache, just persistence. Index and cache have many similarities, but cache of index doesn't make sense. Only `.ark/index` is needed.

kirillt:
Sorry, this PR does not progress. Let's somebody else implement this part.

Issue #24, opened 2022-08-29
Link-to-Web: separate URL and metadata


kirillt:
At the moment, we write all fields in single JSON: URL, title and description; all these fields are treated as parts of resource. This causes `ResourceId` be recalculated every time title or description changes. `ResourceId` should depend only on the URL. The rest must be written into upcoming [metadata storage](https://github.com/ARK-Builders/ARK-Navigator/issues/254).

This changes the very core of ARK Shelf app ([mobile](https://github.com/ARK-Builders/ARK-Shelf) and [desktop](https://github.com/ARK-Builders/ARK-Shelf-Desktop) versions).

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 4 weeks, 1 day from now.
Please review their action plans below:


**1) [ljalbert](https://gitcoin.co/ljalbert) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/29265?mutate_worker_action=approve&worker=ljalbert) | [reject worker](https://gitcoin.co/issue/29265?mutate_worker_action=reject&worker=ljalbert))_.**

Hello, I am an android developer with 10 years experience and have interest working on projects related on mobile and blockchains
I am available to work on this bounty.

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/29265).

gitcoinbot:
Issue Status: 1. Open 2. Started 3. **Submitted** 4. Done 

<hr>

__Work for 5684.7483 ZIL (200.00 USD @ $0.03/ZIL) has been submitted by__:

 1. [@hhio618](https://gitcoin.co/hhio618)



@kirillt please take a look at the submitted work:
* [PR](https://github.com/ARK-Builders/ARK-Navigator/pull/313) by @hhio618

<hr>

* Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/29265)
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/29265).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://discord.gg/gitcoin/'>Gitcoin's Discord</a>
* $1,003,391,691.96 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

kirillt:
[...]

PR #27, opened 2022-09-16
Add link metadata

hhio618:
### Description
#24

### DIFF
--- a/src/link.rs
+++ b/src/link.rs
@@ -5,8 +5,11 @@ use serde::{Deserialize, Serialize};
 use std::hash::{Hash, Hasher};
 use std::path::Path;
 use std::{collections::hash_map::DefaultHasher, fmt};
-use std::{fs::File, io::Write, path::PathBuf};
+use std::{fs, fs::File, io::Write, path::PathBuf};
 use url::Url;
+
+use crate::id::ResourceId;
+const PREVIEWS_RELATIVE_PATH: &str = ".ark/previews";
 
 /// .link File used in ARK Shelf.
 #[derive(Debug, Deserialize, Serialize)]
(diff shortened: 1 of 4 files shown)


kirillt:
Looks like a good set of changes, but unit tests failed:
https://github.com/hhio618/arklib/actions/runs/3066154097

src/link.rs
```diff
@@ -264,14 +265,9 @@ impl OpenGraphTag {
...
-        let mut zip = zip::ZipArchive::new(file.try_clone().unwrap())
-            .expect("Open zip archive");
-        let j_raw = zip
-            .by_name("link.json")
-            .expect("Find link.json in the zip archive");
-
-        let j = serde_json::from_reader(j_raw).expect("Parse link.json");
+        let j_raw = std::fs::read(path).expect("Open link file");
+        let j =
+            serde_json::from_slice(j_raw.as_slice()).expect("Parse link.json");
```
kirillt: Error message is misleading now, am I right better phrase it as "Parse .link file failed"?

src/link.rs
```diff
@@ -113,6 +140,42 @@ impl Link {
...
+        metadata: Vec<u8>,
+        resource_id: ResourceId,
+    ) {
+        let metadata_path = root.as_ref().join(METADATA_RELATIVE_PATH);
+        fs::create_dir_all(metadata_path.to_owned())
+            .expect(&format!("Creating {} directory", METADATA_RELATIVE_PATH));
+        let mut metadata_file = File::create(
+            metadata_path
+                .to_owned()
+                .join(format!("{}", resource_id.crc32)),
```
kirillt, line 173: Actually, we should also use filesize in the path. Otherwise, in case of a collision by CRC-32 we will write into same file 2 different metadata. CRC-32 + filesize is still not perfect, but reduces frequency of collisions significantly without performance penalty.

The best would be to create a method like `to_filename` printing the whole resource id as a string, for usage in filenames.

kirillt:
Superceded by #30

PR #30, opened 2022-11-02
Feat: Metadata storage

hhio618:
### Description
#24 
#27 
https://github.com/ARK-Builders/arklib-android/issues/22
https://github.com/ARK-Builders/ARK-Memo/issues/8

### DIFF
--- a/src/meta.rs
+++ b/src/meta.rs
@@ -3,70 +3,16 @@ use crate::id::ResourceId;
 use anyhow::Error;
 use canonical_path::CanonicalPathBuf;
 use serde::{Deserialize, Serialize};
-use std::ffi::{OsStr, OsString};
-use std::fs::{self, File};
-use std::io::Write;
-use std::path::Path;
 use std::time::SystemTime;
 use walkdir::DirEntry;
-
-const METADATA_RELATIVE_PATH: &str = ".ark/meta";
 
 #[derive(Eq, PartialEq, Hash, Clone, Debug, Deserialize, Serialize)]
 pub struct ResourceMeta {
     pub id: ResourceId,
     pub modified: SystemTime,
-    pub kind: Option<ResourceKind>,
-    pub extra: Option<ResourceExtra>,
 }
 
 impl ResourceMeta {
-    pub fn store<P: AsRef<Path>>(
-        root: P,
-        path: P,
-        resource_id: ResourceId,
-        kind: Option<ResourceKind>,
-        extra: Option<ResourceExtra>,
-    ) {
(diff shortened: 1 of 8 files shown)


src/meta.rs
```diff
@@ -2,21 +2,78 @@ use crate::id::ResourceId;
...
+        kind: Option<ResourceKind>,
+        extra: Option<ResourceExtra>,
+    ) {
+        let metadata_path = root.as_ref().join(METADATA_RELATIVE_PATH);
+        fs::create_dir_all(metadata_path.to_owned())
+            .expect(&format!("Creating {} directory", METADATA_RELATIVE_PATH));
+        let mut metadata_file = File::create(
+            metadata_path
+                .to_owned()
+                .join(format!("{}", resource_id.crc32)),
```
kirillt, line 43: Whole `resource_id` must be used in the filename here.
E.g. `123-45678` where 123 is size of the data, and 45678 is CRC-32 checksum.

src/meta.rs
```diff
@@ -2,21 +2,78 @@ use crate::id::ResourceId;
...
+        )
+        .unwrap();
+        let metadata = Self {
+            id: resource_id,
+            created_date: SystemTime::now(),
+            name,
+            extension,
+            kind,
+            extra,
```
kirillt, line 52: Actually, we only need to store `extra` part. We can put `created_data` into `extra` structure. It's not needed to store `id` (because it is stored in the filename), `name` and `extension` (because it's derived from path which is stored in index). I'm not sure about `kind`, if we can avoid storing it, that would be good. We should store only strictly necessary  data.

src/meta.rs
```diff
@@ -2,21 +2,78 @@ use crate::id::ResourceId;
...
 
-#[derive(Eq, PartialEq, Hash, Clone, Debug)]
+const METADATA_RELATIVE_PATH: &str = ".ark/meta";
+
+#[derive(Eq, PartialEq, Hash, Clone, Debug, Deserialize, Serialize)]
 pub struct ResourceMeta {
     pub id: ResourceId,
-    pub modified: SystemTime,
+    // pub modified: SystemTime,
+    pub created_date: SystemTime,
```
kirillt, line 20: We need `modified` field for tracking when it's necessary to re-scan a resource. It's better to put `created_date` into `extra` for now.

src/meta.rs
hhio618: I think maybe we should keep the id field as it's needed to pass the index tests!

kirillt:
Refactored metadata loading and purified `ResourceMeta` structure. It will be changed further after we merge #29

kirillt (approved):

Issue #38, opened 2023-03-31
Lightweight, single-resource update method


kirillt:
Technically, it is more sound to perform a full update, even when only a single resource is of interest. Because other resources could have been changed, too. However, it is not always easy for the client to process additional updates.

Single-resource `update` function could be convenient. This function should receive path of the resource to update.

Here are some scenarios where this function would be applicable:
1. We could use this function when we delete a resource,  although it would always return `Deleted` event.
2. When an external application is used to modify a resource. In this case, we know that its _id_ has changed, and we want it right in the place where we returned from the external app. For instance, a user browses resources in Gallery, edits an image, and once saving completed, returns to the Gallery. At this moment, either all resources must be updated, or the edited image must be temporarily removed. With the single-resource update, the edited image could simply be updated and presented to the user.

kirillt:
Also, manual update of several selected resources could be useful too.

Examples from Navigator:
1. `ResourcesPresenter:onMoveSelectedResourcesClicked`
2. `ResourcesPresenter:onRemoveSelectedResourcesClicked`

These two methods implement bulk move/remove of resources.

kirillt:
Unit tests should be created verifying that multiple `update_one()` lead to the same state as single `update_all()`.

kirillt:
Implemented in #42

PR #42, opened 2023-08-31
#38: New `update_one` method

kirillt:
New `update_one` method enables library users to update a single entry in the index.

This is useful for scenarios where we performed some action on a resource and want to quickly adjust app model with the index. More details in the issue:
- #38 

The usage in an app should be like this:
1. App has the index. It is allowed to be slightly outdated.
2. User deletes or edits a resource. New resource by the same path has different content.
3. Full updating using `update_all` would spend some time and user could be annoyed by necessity to wait before they can move to next resource in their selection. That's why we call `update_one` and pass path of the edited resource.
4. We receive update from the index and handle it uniformly to any external updates.
5. Since we used this `update_one` method, **the index still can be outdated**, same as in step 1. That's fine because the app model and the index are in sync.
6. When the user is done with their selection, we can call `update_all` and catch up with external updates to the filesystem.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Install Rust
         uses: actions-rs/toolchain@v1
         with:
-            toolchain: nightly
+            toolchain: stable
             override: true
             components: rustfmt, clippy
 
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       - name: Install Rust
         uses: actions-rs/toolchain@v1
         with:
-            toolchain: nightly
```
kirillt, line 19: Double-check that Pdfium works. Nightly was introduced when we were fixing it.

Issue #47, opened 2023-09-21
Declare storage path constants


kirillt:
This repo must be the "single source of truth" for the whole ARK project.

We have also [ArkLib Android](https://github.com/ARK-Builders/arklib-android) repo providing bindings to Rust code for Android and also some other functionality. That extra functionality will go to this repo eventually.

An [important file from the Android repo](https://github.com/ARK-Builders/arklib-android/blob/main/lib/src/main/java/space/taran/arklib/ArkFiles.kt) defines storage path constants:
```
object ArkFiles {
    const val ARK_FOLDER = ".ark"
    const val STATS_FOLDER = "stats"
    const val FAVORITES_FILE = "favorites"

    // User-defined data
    const val TAG_STORAGE_FILE = "user/tags"
    const val SCORE_STORAGE_FILE = "user/scores"
    const val PROPERTIES_STORAGE_FOLDER = "user/properties"

    // Generated data
    const val METADATA_STORAGE_FOLDER = "cache/metadata"
    const val PREVIEWS_STORAGE_FOLDER = "cache/previews"
    const val THUMBNAILS_STORAGE_FOLDER = "cache/thumbnails"
}
```

These constants must be moved to this repo and be imported to the Android repo via bindings.

PR #49, opened 2023-09-30
Define const for differents paths

gwendalF:
#47

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -15,7 +15,7 @@ use anyhow::Error;
 use log;
 
 use crate::id::ResourceId;
-use crate::{INDEX_PATH, STORAGES_FOLDER};
+use crate::{ARK_FOLDER, INDEX_PATH};
 
 #[derive(Eq, Ord, PartialEq, PartialOrd, Hash, Clone, Debug)]
 pub struct IndexEntry {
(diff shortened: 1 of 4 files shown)


src/index.rs
```diff
@@ -802,7 +803,7 @@ mod tests {
 
             assert_eq!(actual.collisions.len(), 0);
             assert_eq!(actual.size(), 2);
-
+            #[cfg(target_os = "unix")]
             file.set_permissions(Permissions::from_mode(0o222))
                 .expect("Should be fine");
```
kirillt, line 808: Curious, what is it about?

kirillt (dismissed):
Almost good, one thing to fix @gwendalF

src/index.rs
gwendalF: It's using the PermissionExt https://doc.rust-lang.org/std/os/unix/fs/trait.PermissionsExt.html which is only available on unix platforms

src/lib.rs
```diff
@@ -21,13 +21,22 @@ use log;
 
 pub const STORAGES_FOLDER: &str = ".ark";
 
-// allowed to be lost (cache)
-pub const INDEX_PATH: &str = "index";
-pub const PREVIEWS_PATH: &str = "previews";
-
 // must not be lost (user data)
 pub const METADATA_PATH: &str = "meta";
 pub const TAGS_PATH: &str = "tags";
```
kirillt, line 26: These two must be replaced by `METADATA_STORAGE_FOLDER` and `TAG_STORAGE_FILE` correspondingly.

src/lib.rs
```diff
@@ -21,13 +21,22 @@ use log;
 
 pub const STORAGES_FOLDER: &str = ".ark";
```
kirillt, line 22: Let's rename to `ARK_FOLDER`.

gwendalF: Done

src/index.rs
kirillt: Ah, I haven't looked at name of the test.. All is clear now.

kirillt (approved):

PR #50, opened 2023-10-01
#48: Atomic files

### LINKED ISSUE #48: Write files atomically
We can't allow half-written files in storages, even in cache.

Write data to a temporary file, then move it to the target path.

gwendalF:
Use atomic files whenever we want to use files. It ensure there are written atomicaly. Also we have the possibility to manage the case when the file is already present and do something with the data. Currently it's just overwritten but something better can be done
Fix #48


src/link.rs
```diff
@@ -27,101 +30,143 @@ impl Link {
...
+        //User defined properties
+        let prop_folder = base_dir
+            .join(PROPERTIES_STORAGE_FOLDER)
+            .join(&id_string);
+        let prop_file = AtomicFile::new(prop_folder)?;
+        modify_json(&prop_file, |data: &mut Option<Properties>| {
+            let properties = self.prop.clone();
+            match data {
+                Some(data) => {
+                    // Hack currently overwrites
```
kirillt, line 109: **Here we should be careful. Better to merge JSONs.**

The easiest way to do it is to allow multiple titles and descriptions :thinking:

src/meta.rs
```diff
@@ -1,45 +1,65 @@
...
+            .join(METADATA_STORAGE_FOLDER)
+            .join(id.to_string()),
+    )?;
+    modify_json(&file, |previous_data: &mut Option<S>| {
+        match previous_data {
+            Some(previous_data) => {
+                // HACK: Find how to handle case where simultaneous writing happens. What is the expected result
+                // Overwrite the data for now
+                *previous_data = metadata.clone();
+            }
```
kirillt, line 34: In case of generated metadata overriding is fine.

src/atomic_file/mod.rs
```diff
@@ -0,0 +1,152 @@
+mod atomic;
+use serde::{de::DeserializeOwned, Serialize};
+use std::io::{Read, Result, Write};
+
+pub use atomic::AtomicFile;
+
+pub fn modify(
```
kirillt, line 7: Is not used. Maybe write a small test case?

src/meta.rs
```diff
@@ -1,30 +1,38 @@
...
-use serde::Serialize;
-
 use crate::id::ResourceId;
-use crate::{ArklibError, Result, ARK_FOLDER, METADATA_STORAGE_FOLDER};
+use crate::{Result, ARK_FOLDER, METADATA_STORAGE_FOLDER};
 
 /// Dynamic metadata: stored as JSON and
 /// interpreted differently depending on kind of a resource
-pub fn store_meta<S: Serialize, P: AsRef<Path>>(
+pub fn store_meta<
```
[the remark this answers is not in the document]
kirillt: Probably safe to delete.

src/atomic_file/atomic.rs
```diff
@@ -0,0 +1,170 @@
...
+    /// well.
+    pub fn compare_and_swap(
+        &self,
+        current: &ReadOnlyFile,
+        new: TmpFile,
+    ) -> Result<()> {
+        let new_path = self.path(current.version + 1);
+        (&new.file).sync_data()?;
+        // May return `EEXIST`.
+        let res = std::fs::hard_link(&new.path, &new_path);
```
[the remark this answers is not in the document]
gwendalF: The new path is indeed coomputed givent the current file_path and adding 1 after it to set the new version

src/atomic_file/atomic.rs
```diff
@@ -0,0 +1,170 @@
...
+    /// If `io::ErrorKind::AlreadyExists` is returned, it means that the latest
+    /// version was not the same as `current` and the operation must be retried
+    /// with a fresher version of the file. Any other I/O error is forwarded as
+    /// well.
+    pub fn compare_and_swap(
+        &self,
+        current: &ReadOnlyFile,
+        new: TmpFile,
+    ) -> Result<()> {
+        let new_path = self.path(current.version + 1);
```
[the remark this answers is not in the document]
gwendalF: It depends of the architecture but on 32 bits machine 4_294_967_295
and 18_446_744_073_709_551_615 on 64 bits machines.
On release mode it will just overflow and restart at 0
Concerning other filename convention, the current implementation used the version number to load the last file. So if it's not the case how to determine the latestt file?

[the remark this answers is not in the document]
kirillt: > It depends of the architecture but on 32 bits machine 4_294_967_295
and 18_446_744_073_709_551_615 on 64 bits machines.

Sounds good, especially this:
> On release mode it will just overflow and restart at 0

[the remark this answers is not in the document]
kirillt: > Concerning other filename convention, the current implementation used the version number to load the last file. So if it's not the case how to determine the latest file?

This is correct and not needed to be changed.

We only need to introduce additional suffix for the filename — _device id_.
It should be:
* Either computed deterministically from the hardware used (fingerprint)
* Or generated once and put somewhere in user settings

Reading:
* Find latest version using incremental numbers (same as now).
* If several versions with same number are found, take the version with local _device id_ and log warning about conflicted versions. These conflicts can come only from remote devices, locally we can't produce conflicts.

src/atomic_file/atomic.rs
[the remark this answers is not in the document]
kirillt: Yes, this is correct if we don't implement pruning. But we don't want to store all versions, because this will increase disk consumption and network sync. We should implement pruning, that means that versions will disappear.

I.e. problematic scenario is like this:
1. We read v2 in function A.
2. Function A does nothing for a while, keeping v2 instance in memory.
3. Function B writes version v3, v4, v5.
4. Versions v2, v3, v4 are automatically deleted because they should never be read again.
5. Function A attempts to write v3. Nobody notices this change because v5 is the latest.

I see 2 ways to solve this:
1. When v3 is written, verify that there are no higher versions (v4 and v5).
2. When v3 is written, verify that v2 exists. Then delete v2.

@gwendalF do you think it's a good idea?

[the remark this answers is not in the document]
gwendalF: We can Indeed check version before making the swap. It all depends on how pruning is done

Cargo.toml
```diff
@@ -32,7 +32,7 @@ tokio = { version = "1", features = ["full"] }
 itertools = "0.10.5"
 once_cell = "1.16.0"
 thiserror = "1"
-
+machine-uid = "0.5.1"
```
kirillt, line 35: Cool crate!

> In Linux, machine id is a single newline-terminated, hexadecimal, 32-character, lowercase ID. When decoded from hexadecimal, this corresponds to a 16-byte/128-bit value. This ID may not be all zeros. This ID uniquely identifies the host. It should be considered “confidential”, and must not be exposed in untrusted environments. And do note that the machine id can be re-generated by root.

Should we hash the id for privacy reasons?

src/atomic_file/atomic.rs
```diff
@@ -92,37 +94,32 @@ pub struct AtomicFile {
...
-                    "`path` must specify a directory name",
-                ));
-            }
+            None => Err(std::io::Error::new(
+                std::io::ErrorKind::InvalidInput,
+                "`path` must specify a directory name",
+            ))?,
         };
-        let prefix = format!("{}.", filename);
+        let prefix = format!("{}_{}.", filename, machine_id);
```
kirillt, line 115: > In Linux, machine id is a single newline-terminated, hexadecimal, 32-character, lowercase ID. When decoded from hexadecimal, this corresponds to a 16-byte/128-bit value. This ID may not be all zeros. This ID uniquely identifies the host. It should be considered “confidential”, and must not be exposed in untrusted environments. And do note that the machine id can be re-generated by root.

We should hash the id for privacy reasons.

src/atomic_file/atomic.rs
```diff
@@ -152,13 +149,22 @@ impl AtomicFile {
     /// version was not the same as `current` and the operation must be retried
     /// with a fresher version of the file. Any other I/O error is forwarded as
     /// well.
+    /// Return the number of old file deleted after swapping
```
kirillt, line 152: "Old version number", right?

kirillt: How this number can be used?

src/atomic_file/atomic.rs
kirillt: Alternative is generating random unique id and saving it into local app data folder.

kirillt:
Do we implement this?

> If several versions with same number are found, take the version with local device id and log warning about conflicted versions. These conflicts can come only from remote devices, locally we can't produce conflicts.

src/prop.rs
```diff
@@ -0,0 +1,92 @@
...
+            .join(ARK_FOLDER)
+            .join(PROPERTIES_STORAGE_FOLDER)
+            .join(id.to_string()),
+    )?;
+    modify_json(&file, |previous_data: &mut Option<S>| {
+        match previous_data {
+            Some(previous_data) => {
+                // HACK: Find how to handle case where simultaneous writing happens. What is the expected result
+                // Overwrite the data for now
+                *previous_data = metadata.clone();
```
kirillt, line 31: Looks like same thing with merging JSONs. We should discuss it.

src/atomic_file/atomic.rs
gwendalF: It's not used. But could be to validate some files are well deleted

src/atomic_file/atomic.rs
mberry: The man pages say to hash it for direct use, so SHA256(machine_id) should be fine.

Otherwise yeah UUID4 works well, there's also UUID7 now, which is also random enough plus time sortable.

mberry: https://www.freedesktop.org/software/systemd/man/machine-id.html

gwendalF: Ok for not using the machine id. However where do you want to store the file containing the Id? Since it's a lib we can pass a variable with the root parameter but don't know which constructor do you want for that. however not sure if we want to pass to the machine identifier on the constructor of the AtomicFile or somewhere else

src/pdf.rs
```diff
@@ -28,8 +28,8 @@ fn initialize_pdfium() {
     )
     .or_else(|_| Pdfium::bind_to_system_library())
     .unwrap();
-    PDFIUM.set(Pdfium::new(bindings)); // Instead of returning the bindings, we
-                                       // cache them in the static initializer
+    let _ = PDFIUM.set(Pdfium::new(bindings)); // Instead of returning the bindings, we
```
mberry, line 31: Perhaps better to do an expect()? Not sure if this can be incorrecly initialised and already set.
```suggestion
    let _ = PDFIUM.set(Pdfium::new(bindings)); // Instead of returning the bindings, we
```

mberry:
Adding a Clippy CI check is good, but I'm getting 1 error here:

https://github.com/ARK-Builders/arklib/blob/4dc4709b1eece7c12112ccb7a90803cad4379785/src/index.rs#L78

It wants the borrow removed

gwendalF:
> Adding a Clippy CI check is good, but I'm getting 1 error here:
> 
> https://github.com/ARK-Builders/arklib/blob/4dc4709b1eece7c12112ccb7a90803cad4379785/src/index.rs#L78
> 
> It wants the borrow removed

We can remove the borrow since it does not appears to be used somewhere else so we can move the path on the open function

src/atomic_file/atomic.rs
```diff
@@ -94,15 +94,16 @@ pub struct AtomicFile {
...
+fn parse_version(filename: Option<&str>) -> Option<usize> {
+    let (_, version) = filename?.rsplit_once('.')?;
     version.parse().ok()
 }
 
 impl AtomicFile {
     pub fn new(path: impl Into<PathBuf>) -> crate::Result<Self> {
         let directory = path.into();
+        // Should assure transfert on internet is safe before sending files containing
+        // this machine_id on the prefix
```
kirillt, line 106: I'll rephrase this comment a bit later.

src/atomic_file/atomic.rs
```diff
@@ -116,14 +117,41 @@ impl AtomicFile {
...
 
-    fn latest_version(&self) -> Result<usize> {
-        let mut max_version = 0;
-        for entry in fs::read_dir(&self.directory)? {
-            if let Some(version) = parse_version(&entry?.file_name()) {
-                max_version = std::cmp::max(max_version, version);
-            }
-        }
-        Ok(max_version)
+    /// Return a vec of files with latest version and the latest version. Multiples files can be found if they comes from different sources.
```
kirillt, line 120: I'll rephrase this comment a bit later.

src/atomic_file/atomic.rs
```diff
@@ -116,14 +117,41 @@ impl AtomicFile {
...
+    /// For example one from cellphone and one from computer can both a a version 2.
+    fn latest_version(&self) -> Result<(Vec<ReadOnlyFile>, usize)> {
+        let files_iterator = fs::read_dir(&self.directory)?.flatten();
+        let (files, version) = files_iterator.into_iter().fold(
+            (vec![], 0),
+            |(mut files, mut max_version), entry| {
```
kirillt, line 126: Let's call it `curr_max_version` or something like this. Can be confused with `MAX_VERSION_FILES` constant.

src/atomic_file/atomic.rs
```diff
@@ -116,14 +117,41 @@ impl AtomicFile {
...
+                    // It's possible to have same version for two files coming from different machines
+                    // Add this files to the result
+                    if version >= max_version {
+                        let read_only = ReadOnlyFile {
+                            version,
+                            path: entry.path(),
+                        };
+                        files.push(read_only);
+                        max_version = version;
+                    }
```
kirillt, line 138: Does it really work correctly in case of conflicts for several versions?
E.g. versions: `ver1_a`, `ver2_a`, `ver3_a`, `ver3_b`, `ver4_a`, `ver4_b`, `ver5_a`.

It also can depend on the order we traverse version files, different order would give different results.

src/prop.rs
```diff
@@ -26,8 +26,7 @@ pub fn store_properties<
     modify_json(&file, |previous_data: &mut Option<S>| {
         match previous_data {
             Some(previous_data) => {
-                // HACK: Find how to handle case where simultaneous writing happens. What is the expected result
-                // Overwrite the data for now
+                // Overwrites data
```
kirillt, line 29: No no, in this place we shouldn't overwrite.

We can and should overwrite plain metadata, but in this place we work with properties, i.e. user data.

src/prop.rs
```diff
@@ -26,8 +26,7 @@ pub fn store_properties<
...
                 *previous_data = metadata.clone();
```
kirillt, line 30: Let's renamed `metadata` into `properties` or `prop` here. We need to merge older and newer version.

The format for merge actually can look weird at first. I propose to use a list for each field.
E.g. `{ title: ["cool website", "amazing website"], description: [ "just one" ] }`

Merging should keep only unique values, i.e. field `description: ["just one", "just one"]` is not valid.

In UI, we'll display only 1 element of each list and hide the rest with some warning icon, so user would be able to review conflicting fields.

See https://github.com/ARK-Builders/ark-components-android/issues/32

src/atomic_file/atomic.rs
GwendalFernet: @kirillt The order of parsing the files should not change anything. The list is filter after. I'll add some test

src/prop.rs
GwendalFernet: @kirillt Ok for not overwriting, but what's the behavior you want

src/prop.rs
GwendalFernet: @kirillt So every properties should be a list and we don't accept plain string for example?

gwendalF:
I've updated the case to not override properties. May need check because there is a bunch of clone and staff, so we can surely do it better

src/atomic_file/atomic.rs
```diff
@@ -309,10 +310,58 @@ mod tests {
...
+
+    #[rstest]
+    #[case(3, &[1, 3], "case_1")]
+    #[case(5, &[2, 4], "case_2")]
+    #[case(10, &[3, 5, 7, 9, 10], "case_3")]
+    #[case(15, &[5, 14, 15], "case_4")]
+    fn latest_version(
+        #[case] versions: usize,
+        #[case] cellphone_versions: &[usize],
+        #[case] temp_name: &str,
```
kirillt, line 328: _(No issue here)_ I should learn about `rstest`.

src/atomic_file/atomic.rs
```diff
@@ -309,10 +310,58 @@ mod tests {
...
+    ) {
+        // Create the files without atmic to handles files names
+        // let versions = 10;
+        // let cellphone_versions = &[3, 5, 7, 9, 10];
+        let dir = TempDir::new(temp_name).unwrap();
+        let root = dir.path();
+        // let root = PathBuf::from("./latest_version");
+        let file = AtomicFile::new(&root).unwrap();
```
kirillt, line 336: Can we move this line closer to its usage? I.e. right before `assert_eq!(file.latest_version().unwrap().1, versions)`

src/atomic_file/atomic.rs
```diff
@@ -309,10 +310,58 @@ mod tests {
...
+        #[case] versions: usize,
+        #[case] cellphone_versions: &[usize],
+        #[case] temp_name: &str,
+    ) {
+        // Create the files without atmic to handles files names
+        // let versions = 10;
+        // let cellphone_versions = &[3, 5, 7, 9, 10];
+        let dir = TempDir::new(temp_name).unwrap();
+        let root = dir.path();
+        // let root = PathBuf::from("./latest_version");
```
kirillt, line 335: Commented-out code should be removed before merging.

kirillt:
@gwendalF Thanks, good changes. Could you tell more about visitor pattern here?
> May need a visitor pattern to merge two serde_json::Value

src/atomic_file/atomic.rs
```diff
@@ -0,0 +1,369 @@
...
+        };
+        let prefix = format!("{}_{}.", filename, machine_id);
+        Ok(Self { directory, prefix })
+    }
+
+    /// Return the latest version together with vector of the
+    /// files matching this version. Multiple files for the same version
+    /// can appear due to usage of file syncronization. Different devices
+    /// can create same version simultaneously.
+    fn latest_version(&self) -> Result<(Vec<ReadOnlyFile>, usize)> {
```
kirillt, line 124: _(Nitpick)_ Putting version first in the returned tuple would be more readable since the name of the function is explicit about it.

src/link.rs
```diff
@@ -27,101 +30,142 @@ impl Link {
...
+                    *data = properties;
+                }
+                None => *data = Some(properties),
+            }
+        })?;
```
kirillt, line 113: It looks like we should use `fn store_properties` here.

We also might need to change `struct Properties` declaration.

src/link.rs
```diff
@@ -27,101 +30,142 @@ impl Link {
...
+                let graph = data.clone();
+                println!("Trying to save: {with_preview} with {graph:?}");
+                match file_data {
+                    Some(file_data) => {
+                        // Hack currently overwrite
+                        *file_data = graph;
+                    }
+                    None => *file_data = Some(graph),
+                }
+            })?;
```
kirillt, line 131: Probably we should create `meta.rs` for handling generated metadata now. It would be similar to `prop.rs` but always overriding versions, because it's just a cache.

src/atomic_file/atomic.rs
kirillt: - #54
- #55

Cargo.toml
gwendalF: So unfortunately it doesn't take darwin architecture into account. I'll make a PR on their crate but don't knwo when we should be able to use it for ARM users

src/atomic_file/atomic.rs
gwendalF: The order should have no influence of the results since only files with latest version are returned

src/prop.rs
gwendalF: A merging function has been added and used for properties

src/atomic_file/atomic.rs
gwendalF: It is needed to get the prefix of the file. But I defined it just before its usage

src/link.rs
gwendalF: Is there reason to not always use Vec<String> ? It would easy the process of deserialisation and allow to define a uniq type of properties instead of having either String or Vec<String>

src/link.rs
```diff
@@ -27,101 +31,129 @@ impl Link {
...
+    ) -> Result<Properties> {
+        let path = root
+            .as_ref()
+            .join(ARK_FOLDER)
+            .join(PROPERTIES_STORAGE_FOLDER)
+            .join(id.to_string());
+        let file = AtomicFile::new(path)?;
+        let current = file.load()?;
+        let data = current.read_to_string()?;
+        let user_meta: Properties = serde_json::from_str(&data)?;
```
kirillt, line 54: Am I right, `data` can be array here?

kirillt:
I've realized that we should use lightweight version of atomic files for resources.

In most cases, there will be no other resource with same id. The could be collisions but they should not be treated as versions. Probably simple "write in temporary location, move to the final path" approach should be implemented, too? @gwendalF what do you think?

src/link.rs
gwendalF: The struct Can have properties as Vector. 
Is it an issue if I update the definition of Properties?

kirillt:
Should we write previews atomically, too?

gwendalF:
@kirillt writing to temp and move is kind of the same of what atomic file is doing. But without versioning moving would erase previous file. So it will only solve half written files but not synchronisation.
Half written could only happens if the process is killed during writing.

Can be done but I don't know when half written files happens

src/storage/meta.rs
```diff
@@ -0,0 +1,93 @@
...
+                // not be used on synced devices.
+            }
+            None => *current_meta = Some(new_meta),
+        }
+    })?;
+    Ok(())
+}
+
+/// The file must exist if this method is called
+pub fn load_raw_metadata<P: AsRef<Path>>(
```
kirillt, line 41: I'm not sure where this should be used. @gwendalF Could you check?

Cargo.toml
kirillt: @Rizary might be interesting to you, too

kirillt:
@gwendalF yeah, the idea behind atomics for resources is that we just want to protect from half-written files because it should be considered a unique case when we write a new resource. Of course, there are hash collisions sometimes, but we can use some nonce (maybe just timestamp) in the filename to protect from that. The scenario with editing properties is much more prone to dirty writes because it can be done from different devices multiple times. If we want to create same resource again, we just check if it exists already or not.

gwendalF:
@kirillt What resource are you talking about? Properties?

gwendalF:
@kirillt  But the resource (link) has some properties. So writing properties is needed when we create a link

kirillt:
@gwendalF yes sure, we just should use different means to write metadata and content itself. Because property and other storages are shared between multiple resources. If something goes wrong during resource write, only this resource will be lost. If something foes wrong while writing a storage, then we risk losing metadata for other resources too.

For the properties storage it's not critical though because it's folder-based, but the tag storage is a single file containing multiple entries.

kirillt:
Just a bit about terminology: "resource" means some user's content. We don't call it file because if we change its path, the resource still remains the same.

We have various "metadata" associated with resources: 1) user-defined, e.g. tags, properties, scores etc. 2) app-generated, i.e. cache: metadata/kinds and previews/thumbnails

gwendalF:
@kirillt 
It's not clear for me what you wish. `I've realized that we should use lightweight version of atomic files for resources.` -> What resources? Link? For example tag storage in never used in the current arklib, only the path is defined.
If you wish to write Link without atomic, it's possible. But then what about properties? And what about metadata ? And previews?

kirillt:
@gwendalF a "resource" is any user's file. In Shelf app, we work only with Link resources, but Navigator works with any resources: images, videos, documents. We store resources as files in a folder selected by the user.

As I said before, properties and other user-defined metadata (tags, scores) should use atomic files you've implemented, i.e. it shouldn't be changed.

Resources, i.e. any file outside `.ark` subfolder, should use simple write-tmp-and-move approach.

Previews, thumbnails and generated metadata don't really require atomic writing, because we always can regenerate them. But we can also use write-tmp-and-move approach.

gwendalF:
@kirillt I've updated the Link to use your solution of tempfile and moving to destination

src/atomic/file.rs
```diff
@@ -293,33 +293,37 @@ mod tests {
...
         (&temp)
-            .write_all(other_machine.as_bytes())
+            .write_all(content_remote.as_bytes())
             .unwrap();
         file.compare_and_swap(&current, temp).unwrap();
+
         let version_2_path = file.path(2);
         let rename_path =
             root.join(format!("{}_cellphoneId.1", root.display()));
         fs::rename(version_2_path, rename_path).unwrap();
```
kirillt, line 321: Not sure what exactly we are testing here, will look closer a bit later.

src/link.rs
```diff
@@ -25,6 +25,20 @@ pub struct Properties {
     pub title: String,
     pub desc: Option<String>,
 }
+/// Write data to a tempory file and move that written file to destination
+///
+/// May failed if writing or moving failed
+fn temp_and_move(
```
kirillt, line 31: Probably we should expose this method so apps could use it directly (e.g. Retouch or Memo).

kirillt: We should consider making "atomic files" API public, too. But not sure about the use-case yet.

Cargo.toml
kirillt: https://github.com/ARK-Builders/arklib/issues/61

kirillt (approved):
Good job

PR #51, opened 2023-10-01
Add error for different use case

gwendalF:
Define custom errors type. More type could be use if needed and remove the anyhow completely. Anyhow is better for application developper who want to manage every errors. For a library it's better to provide different errors and let the consumer decide what to do
https://github.com/ARK-Builders/arklib/issues/43

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,0 +1,33 @@
+use std::str::Utf8Error;
+
+use thiserror::Error;
+
+pub type Result<T> = std::result::Result<T, ArklibError>;
+
+#[derive(Error, Debug)]
+pub enum ArklibError {
+    #[error("IO error")]
+    Io(#[from] std::io::Error),
+    #[error("Invalid path for ressource")]
+    Path,
+    #[error("There is some collision: {0}")]
+    Collision(String),
+    #[error("Parsing error")]
+    Parse,
+    #[error("Networking error")]
+    Network,
+    #[error(transparent)]
+    Other(#[from] anyhow::Error),
+}
+
+impl From<reqwest::Error> for ArklibError {
+    fn from(_: reqwest::Error) -> Self {
+        Self::Network
+    }
+}
+
+impl From<Utf8Error> for ArklibError {
+    fn from(_: Utf8Error) -> Self {
+        Self::Parse
+    }
+}
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -8,8 +8,8 @@ pub type Result<T> = std::result::Result<T, ArklibError>;
 pub enum ArklibError {
     #[error("IO error")]
     Io(#[from] std::io::Error),
-    #[error("Invalid path for ressource")]
-    Path,
+    #[error("Path error: {0}")]
+    Path(String),
     #[error("There is some collision: {0}")]
     Collision(String),
     #[error("Parsing error")]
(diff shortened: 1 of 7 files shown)


src/errors.rs
```diff
@@ -0,0 +1,33 @@
...
+
+use thiserror::Error;
+
+pub type Result<T> = std::result::Result<T, ArklibError>;
+
+#[derive(Error, Debug)]
+pub enum ArklibError {
+    #[error("IO error")]
+    Io(#[from] std::io::Error),
+    #[error("Invalid path for ressource")]
```
kirillt, line 11: Resource with one `s` :wink:  I already fixed that once, so now telling in comments..

src/index.rs
```diff
@@ -90,14 +89,16 @@ impl ResourceIndex {
...
+                            str.parse().map_err(|_| ArklibError::Parse)?,
+                        ))
                     };
 
                     let id: ResourceId = {
-                        let str = parts.next().ok_or(Error::msg(
+                        let str = parts.next().ok_or(anyhow::Error::msg(
                             "Couldn't find second part of index entry",
                         ))?;
                         ResourceId::from_str(str)?
```
kirillt, line 104: But why `anyhow` here?

src/index.rs
```diff
@@ -386,7 +381,9 @@ impl ResourceIndex {
                             }
 
                             // the caller must have ensured that the path was updated
-                            return Err(Error::msg("Nothing to update"));
+                            return Err(ArklibError::Collision(
+                                "Nothing to update".into(),
```
kirillt, line 385: Could you also change the message to `New content has the same id.`

src/index.rs
```diff
@@ -515,17 +514,14 @@ fn discover_paths<P: AsRef<Path>>(
...
     if metadata.is_dir() {
-        return Err(Error::msg("DirEntry is directory"));
+        return Err(ArklibError::Path);
     }
 
     let size = metadata.len();
     if size == 0 {
-        return Err(Error::msg("Empty resource"));
+        return Err(ArklibError::Path);
     }
```
kirillt, line 525: Looks like a problem with the content, not with the path.

src/index.rs
```diff
@@ -515,17 +514,14 @@ fn discover_paths<P: AsRef<Path>>(
...
 }
 
-fn scan_entry(
-    path: &CanonicalPath,
-    metadata: Metadata,
-) -> Result<IndexEntry, Error> {
+fn scan_entry(path: &CanonicalPath, metadata: Metadata) -> Result<IndexEntry> {
     if metadata.is_dir() {
-        return Err(Error::msg("DirEntry is directory"));
+        return Err(ArklibError::Path);
```
kirillt, line 519: Detailed message would be good unless that's the only `ArklibError::Path`  in the whole codebase.

src/index.rs
gwendalF: I fix it using Parse Error seems it's more like a parsing issue

src/index.rs
gwendalF: I updated the message

src/index.rs
gwendalF: I used std::io::Error::new(std::io::Errorkind::InvalidData

src/index.rs
gwendalF: I have added a String message on the Path error

kirillt (approved):

Issue #41, opened 2023-08-31
Ensure that `update_all` and `update_one` are consistent


kirillt:
- [ ] Unit tests covering `update_one`
- [ ] Unit tests covering `update_all`

Then, we want to verify that:
* `update_all` can be replaced by multiple `update_one`

- [ ] Unit tests for simple cases with both `update_all` and `update_one`
- [ ] Randomized test scenario checking different combinations

mberry:
The issue I see arising with `update_one` is that it takes a new path and old id, but also uses the path as a key with `self.path2id[path]` before updating the index. 
Think some of the functions need to be run in a different order, like the collision check at the end before returning

PR #52, opened 2023-10-14
#41: `update_one` tests

oluiscabral:
- #41

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -904,4 +904,23 @@ mod tests {
         assert!(new2 > old2);
         assert!(new2 > new1);
     }
+
+    #[test]
+    fn should_correctly_update_one_resource() {
+        run_test_and_clean_up(|path| {
+            create_file_at(path.clone(), Some(FILE_SIZE_1), Some(FILE_NAME_1));
+            create_file_at(path.clone(), Some(FILE_SIZE_2), Some(FILE_NAME_2));
+            let mut actual = ResourceIndex::build(path.clone());
+            let mut file_path = path.clone();
+            file_path.push(FILE_NAME_1);
+            let old_id = ResourceId {
+                data_size: FILE_SIZE_1,
+                crc32: CRC32_1,
+            };
+            let path_buf = CanonicalPathBuf::canonicalize(&file_path).unwrap();
+            let update = actual
+                .update_one(path_buf.clone(), old_id)
+                .expect("Should update one resource correctly");
+        });
+    }
 }


src/index.rs
```diff
@@ -904,4 +904,23 @@ mod tests {
...
+            let mut actual = ResourceIndex::build(path.clone());
+            let mut file_path = path.clone();
+            file_path.push(FILE_NAME_1);
+            let old_id = ResourceId {
+                data_size: FILE_SIZE_1,
+                crc32: CRC32_1,
+            };
+            let path_buf = CanonicalPathBuf::canonicalize(&file_path).unwrap();
+            let update = actual
+                .update_one(path_buf.clone(), old_id)
```
kirillt, line 922: Now I'm curious what happens if a user of the lib would pass wrong `old_id`...

src/index.rs
```diff
@@ -904,4 +904,23 @@ mod tests {
...
+                .expect("Should update one resource correctly");
+        });
```
kirillt, line 924: In this current scenario, the update should be empty (nothing changed).

Probably reasoning with paths instead of files would be better here:
* If we remove a file with id `id1` and path `path1` then the update is `deleted(id1)`
* If we modify a file by path `path1` so it's new id is `id2` then the update is `deleted(id1), added(id2)`
* If we rename/move a file with `id1` from `path1` to `path2` then the update is `deleted(id1), added(id1)`
Please, double-check.

src/index.rs
```diff
@@ -904,4 +904,23 @@ mod tests {
         assert!(new2 > old2);
         assert!(new2 > new1);
     }
+
```
kirillt, line 907: Idea for a test case:
1. Build the index of folder with paths e.g. `root/A` and `root/B`.
2. Create file by path `root/C`.
3. Call `update_one("root/C", id)`.
Two cases:
a) `id` matches the content of new resource
b) `id` does not 
4. What should be the result?

src/index.rs
kirillt: _Notion of a "file" is confusing here because it is combination of path and content, but index works with paths and contents separately. A "resource" is an identified content, a "path" it's current location._

src/index.rs
oluiscabral: Nice, alright. Should it replace the current one or is it a new unittest?

kirillt: New one

kirillt:
Should we create new struct for the error `Collision("New content has the same id")`? Word "Collision" can be confusing here since it can be caused not only by hash collision but also by wrong caller's code. The idea is that `update_one` is a quick way to update index without checking all resources in situations when the caller knows for sure that it was changed. E.g. when the caller edits the resource.

So, if the caller doesn't have information that the resource changed, it's a mistake to call this method. But if the resource was changed indeed (timestamp must be changed, too) and new content has same size and same CRC-32 hash — then it's a collision indeed.

We can also split single error into two different errors for clarity, depending on was the timestamp changed or not.

kirillt:
Superseded by #72

PR #53, opened 2023-10-16
Replaced id2path with a function call

mberry:
Putting this out there as an alternative to the current struct. The resource lists are duplicated, but also get out of sync with collisions. This removes id2path and replaces all current uses with path2id. For future use id2path is also now a function call.

* Modified all previous key searches using it.
* Added some function doc notes
* Fixed target_os -> target_family for unix file permissions 
* Removed a few noop test assertions

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -24,9 +24,7 @@ pub struct IndexEntry {
 
 #[derive(Clone, Debug)]
 pub struct ResourceIndex {
-    pub id2path: HashMap<ResourceId, CanonicalPathBuf>,
     pub path2id: HashMap<CanonicalPathBuf, IndexEntry>,
-
     pub collisions: HashMap<ResourceId, usize>,
     root: PathBuf,
 }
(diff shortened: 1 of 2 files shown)


src/index.rs
```diff
@@ -24,9 +24,7 @@ pub struct IndexEntry {
 
 #[derive(Clone, Debug)]
 pub struct ResourceIndex {
-    pub id2path: HashMap<ResourceId, CanonicalPathBuf>,
```
kirillt, line 27: Probably we should keep it but make it private. The `id2path` mapping removes necessity to create a cache for the `find(|(_, v)| v.id == *id)` computation. The trade-off is that the update logic is more complicated, but the `main` version looks more performant.

mberry: Fair enough. Found it a bit confusing for what is happening as they both have the similar names but the collections get out of sync quickly.

How to check performance?

Also I forgot to run cargo fmt, which is why the build failed :/

kirillt: > Found it a bit confusing for what is happening as they both have the similar names but the collections get out of sync quickly.

Probably we should introduce some `fn check_consistency()` and use it in unit tests.

> How to check performance?

Oh yeah, we need some benchmarks..

Issue #34, opened 2023-03-06
Missing file causes index loading failure and leads to complete rebuild


kirillt:
1. Build index of some folder using `ark-cli monitor`.
2. Verify `.ark/index` exists.
3. Remove a file from the folder. Verify the index has entry with its path.
4. Load index.
Last step fails, the index is completely rebuilt in this case.

Pay attention to `WARN` lines:
```
[kirill@lenovo TEST]$ RUST_LOG=info ark-cli monitor
Building index of folder /tmp/TEST
[2023-03-06T16:21:33Z INFO  arklib] Index has not been registered before
[2023-03-06T16:21:33Z INFO  arklib::index] Loading the index from file
[2023-03-06T16:21:33Z WARN  arklib::index] No persisted index was found by path /tmp/TEST/.ark/index
[2023-03-06T16:21:33Z INFO  arklib::index] Building the index from scratch
[2023-03-06T16:21:36Z INFO  arklib::index] Index built
[2023-03-06T16:21:36Z INFO  arklib::index] Storing the index to file
[2023-03-06T16:21:36Z INFO  arklib] Index was registered
Build succeeded in 3.667931346s

Updating succeeded in 158.747289ms

^C
[kirill@lenovo TEST]$ RUST_LOG=info ark-cli monitor     #just load index from the file and check for updates
Building index of folder /tmp/TEST
[2023-03-06T16:21:57Z INFO  arklib] Index has not been registered before
[2023-03-06T16:21:57Z INFO  arklib::index] Loading the index from file
[2023-03-06T16:21:58Z INFO  arklib::index] Storing the index to file
[2023-03-06T16:21:58Z INFO  arklib] Index was registered
Build succeeded in 297.370813ms

^C
[kirill@lenovo TEST]$ grep gagarin .ark/index
1678115269692 200433-880886451 gagarin.jpg
[kirill@lenovo TEST]$ mv gagarin.jpg /tmp/

[kirill@lenovo TEST]$ RUST_LOG=info ark-cli monitor     #must load the index and remove disappeared resource
Building index of folder /tmp/TEST
[2023-03-06T16:22:43Z INFO  arklib] Index has not been registered before
[2023-03-06T16:22:43Z INFO  arklib::index] Loading the index from file
[2023-03-06T16:22:43Z WARN  arklib::index] No such file or directory (os error 2)
[2023-03-06T16:22:43Z INFO  arklib::index] Building the index from scratch
[...]

PR #56, opened 2023-10-30
#34: Don't return early when loading index, jus skip missing data

gwendalF:
The issue caused when deleting an entry is that the parsing failed. I propose to just skip line with missing data
https://github.com/ARK-Builders/arklib/issues/34

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -82,47 +82,21 @@ impl ResourceIndex {
             root: root_path.clone(),
         };
 
-        // We should not return early in case of errors
+        // We should not return early in case of missing files
         for line in BufReader::new(file).lines() {
             if let Ok(entry) = line {
                 let mut parts = entry.split(' ');
 
-                let modified: SystemTime = {
-                    match parts.next() {
-                        Some(millis) => match millis.parse() {
-                            Ok(millis) => {
-                                UNIX_EPOCH.add(Duration::from_millis(millis))
-                            }
-                            Err(_) => {
-                                log::warn!("Can't parse {millis} to u64");
-                                continue;
-                            }
-                        },
-                        None => {
-                            log::warn!("Missing timestamp from {entry}");
-                            continue;
-                        }
-                    }
+                let modified = {
+                    let str = parts.next().ok_or(ArklibError::Parse)?;
+                    UNIX_EPOCH.add(Duration::from_millis(
+                        str.parse().map_err(|_| ArklibError::Parse)?,
+                    ))
                 };
 
-                let id: ResourceId = {
-                    match parts.next() {
-                        Some(str) => match ResourceId::from_str(str) {
-                            Ok(id) => id,
-                            Err(_) => {
-                                log::warn!(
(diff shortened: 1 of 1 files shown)


src/index.rs
```diff
@@ -83,29 +82,62 @@ impl ResourceIndex {
...
-                    UNIX_EPOCH.add(Duration::from_millis(
-                        str.parse().map_err(|_| ArklibError::Parse)?,
-                    ))
+                    match parts.next() {
+                        Some(millis) => match millis.parse() {
+                            Ok(millis) => {
+                                UNIX_EPOCH.add(Duration::from_millis(millis))
+                            }
+                            Err(_) => {
+                                log::warn!("Can't parse {millis} to u64");
```
kirillt, line 97: I think we should not continue the loop. If the file is broken, reading it makes no sense.

src/index.rs
```diff
@@ -83,29 +82,62 @@ impl ResourceIndex {
...
+                                continue;
+                            }
+                        },
+                        None => {
+                            log::warn!("Missing timestamp from {entry}");
```
kirillt, line 102: Same here
> I think we should not continue the loop. If the file is broken, reading it makes no sense.

src/index.rs
```diff
@@ -83,29 +82,62 @@ impl ResourceIndex {
...
-                    let str = parts.next().ok_or(ArklibError::Parse)?;
-                    ResourceId::from_str(str)?
+                    match parts.next() {
+                        Some(str) => match ResourceId::from_str(str) {
+                            Ok(id) => id,
+                            Err(_) => {
+                                log::warn!(
+                                    "Can't parse {str} to a RessourceId"
+                                );
+                                continue;
```
kirillt, line 116: Same here
> I think we should not continue the loop. If the file is broken, reading it makes no sense.

src/index.rs
```diff
@@ -83,29 +82,62 @@ impl ResourceIndex {
...
+                            }
+                        },
+                        None => {
+                            log::warn!(
+                                "Missing ressource id part on the line {entry}"
+                            );
+                            continue;
```
kirillt, line 123: Same here
> I think we should not continue the loop. If the file is broken, reading it makes no sense.

src/index.rs
```diff
@@ -83,29 +82,62 @@ impl ResourceIndex {
...
-                index.insert_entry(path, IndexEntry { id, modified });
+                match CanonicalPathBuf::canonicalize(&path) {
+                    Ok(path) => {
+                        log::trace!("[load] {} -> {}", id, path.display());
+                        index.insert_entry(path, IndexEntry { id, modified });
+                    }
+                    Err(_) => {
+                        log::warn!("File {} not found", path.display());
+                        continue;
+                    }
```
kirillt, line 139: Here continuing should be good :+1:

gwendalF: Ok I've updated to return early if the file can't be parsed

kirillt (approved):

PR #57, opened 2023-11-04
#45: Change the parameter's type of `update_one` to generic `Path`

### LINKED ISSUE #45: Function `update_one` cannot accept `CanonicalPathBuf` for deleted files
This is an oversight of implementation in #42

See #38 for more context. **TL;DR:** We use `update_one` for cases when a resource by some path changed or was deleted at all. We can't canonicalize a non-existent path, so we fail to call `update_one` in this case. This API should receive plain `Path`-like type: `Path`, `PathBuf` or be generic with `AsRef<Path>`.

Rizary:
fix #45 by changing the update_one parameter to `Asref<Path>`


src/index.rs
```diff
@@ -340,38 +340,47 @@ impl ResourceIndex {
...
         log::debug!("Updating a single entry in the index");
-
-        let path = path_buf.as_canonical_path();
+        
+        // the caller must have ensured that the path is exist
+        if !path.as_ref().exists() {
+            return Err(ArklibError::Path(
+                "Path does not exist".into(),
+            ));
+        }
```
kirillt, line 353: We actually need to support this case as normal scenario. When we delete a resource, we need to update the index. Consumer app first deletes the resource then calls `update_one` with its path, which already doesn't exist.

Cargo.lock
kirillt, line 1: Probably should not be checked in. Because the consumer code may need to specify its own versions of crates.

src/index.rs
```diff
@@ -340,38 +340,47 @@ impl ResourceIndex {
 
     pub fn update_one(
         &mut self,
-        path_buf: CanonicalPathBuf,
+        path: &dyn AsRef<Path>,
```
kirillt, line 343: :+1:

Cargo.lock
kirillt: @gwendalF is it correct?

src/index.rs
Rizary: noted.

src/index.rs
```diff
@@ -340,61 +340,101 @@ impl ResourceIndex {
...
+                    Some(k)
+                } else {
+                    None
+                }
+            })
+            .collect_vec();
+        for p in old_path {
+            self.path2id.remove(&p);
+        }
+        self.id2path.remove(&old_id);
```
kirillt, line 428: Probably we should handle collisions differently to remove necessity iterate through `path2id`. But this should be implemented as separate issue.

kirillt (approved):

src/index.rs
kirillt: - #58

Issue #54, opened 2023-10-17
Better privacy with random unique device ids


kirillt:
We introduce [machine-uid](https://docs.rs/machine-uid/latest/machine_uid/) crate in this PR:
- #50 

The crate comes with a disclaimer:
> In Linux, machine id is a single newline-terminated, hexadecimal, 32-character, lowercase ID. When decoded from hexadecimal, this corresponds to a 16-byte/128-bit value. This ID may not be all zeros. This ID uniquely identifies the host. It should be considered “confidential”, and must not be exposed in untrusted environments. And do note that the machine id can be re-generated by root.

Alternative would be generating a random device id, storing it in app data folder, sharing it privately with other devices when necessary, etc.

We've decided in favor of machine-uid because it's much easier to implement and outside entity can't figure out the ids if secure transport is utilized. But with unique random ids, unencrypted transport could be used. If privacy becomes a concern, we should implement this approach.

kirillt:
Unique ids per app instance might be preferred over `machine-uid`. It provides better privacy comparing to fingerprinting, because app id is different for each installation and can be reset. After resetting versions marked with old id will look foreign for the app though, but this is probably not an issue.

For implementing unique ids, we need to implement some kind of callbacks into arklib which pass location of app data, this location is different for Desktop and Mobile versions (Android, and later iOS). It means, apps will need to pass the location to arklib during loading.

kirillt:
Unique ids per app instance might be preferred over `machine-uid`. It provides better privacy comparing to fingerprinting, because app id is different for each installation and can be reset. After resetting versions marked with old id will look foreign for the app though, but this is probably not an issue.

For implementing unique ids, we need to implement some kind of callbacks into arklib which pass location of app data, this location is different for Desktop and Mobile versions (Android, and later iOS). It means, apps will need to pass the location to arklib during loading.

kirillt:
Also, `machine-uid` is not supported on Android yet.

PR #64, opened 2024-01-04
#54: Replace `machine-id` with stored `uuid`

alvinosh:
- #54

### DIFF
--- a/src/atomic/file.rs
+++ b/src/atomic/file.rs
@@ -4,7 +4,7 @@ use std::io::{Error, ErrorKind, Read, Result};
 use std::os::unix::fs::MetadataExt;
 use std::path::{Path, PathBuf};
 
-use crate::id::device_id;
+use crate::id::app_id;
 
 const MAX_VERSION_FILES: usize = 10;
 
@@ -48,8 +48,8 @@ impl Drop for TmpFile {
 
 #[derive(Clone)]
 pub struct ReadOnlyFile {
-    version: usize,
-    path: PathBuf,
+    pub version: usize,
+    pub path: PathBuf,
 }
 
 /// This struct is the only way to read the file. Both path and version are private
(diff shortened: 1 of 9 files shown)


src/lib.rs
```diff
@@ -44,12 +49,24 @@ pub const METADATA_STORAGE_FOLDER: &str = "cache/metadata";
 pub const PREVIEWS_STORAGE_FOLDER: &str = "cache/previews";
 pub const THUMBNAILS_STORAGE_FOLDER: &str = "cache/thumbnails";
 
+pub const APP_ID_FILE: &str = "device";
```
kirillt, line 52: Let's use some value like `"app_id"` for consistency.

src/lib.rs
```diff
@@ -44,12 +49,24 @@ pub const METADATA_STORAGE_FOLDER: &str = "cache/metadata";
...
         RwLock::new(HashMap::new());
 }
+lazy_static! {
+    pub static ref APP_ID_PATH: RwLock<Option<PathBuf>> = RwLock::new(None);
+}
+
+pub fn initialize() {
+    INIT.call_once(|| {
+        log::info!("Initializing arklib");
+        app_id::load("./").unwrap();
```
kirillt, line 67: Aha.. that's interesting. Might make it easier on any platform.

src/id.rs
```diff
@@ -108,16 +108,76 @@ impl ResourceId {
     }
 }
 
+pub mod app_id {
```
kirillt, line 111: Is it worth to put it into separate file `app_id`?

src/atomic/file.rs
```diff
@@ -105,7 +107,8 @@ impl AtomicFile {
         // This UID must be treated as confidential information.
         // Depending on network transport used to sync the files (if any),
         // it can leak to an unauthorized party.
-        let machine_id = machine_uid::get()?;
+        let machine_id = app_id::read()?;
```
kirillt, line 110: We don't have to keep this term "machine id", we can call the variable something like `local_peer` to reflect the purpose of having it.

kirillt (approved):
Simple and clean, good job.

PR #71, opened 2024-01-15
make unit tests of update_one and update_all method

funnyVOLT:
- Covered `update_one` unit test
- Covered `update_all` unit test
In the third key point, you mentioned the potential substitution of `update_all` with multiple instances of the `update_one` method. 
`update_one` method have some problems. 
For instance, when the first parameter of the `update_one` method is the path of a newly added file, it returns an error. 
Furthermore, the same holds true when the first parameter of the `update_one` method corresponds to a file that has not been updated.
To leverage the `update_one` method in a manner similar to the `update_all` method, it is imperative to refine and update the `update_one` method accordingly.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -30,7 +30,7 @@ pub struct ResourceIndex {
     root: PathBuf,
 }
 
-#[derive(Debug)]
+#[derive(PartialEq, Debug)]
 pub struct IndexUpdate {
     pub deleted: HashSet<ResourceId>,
     pub added: HashMap<CanonicalPathBuf, ResourceId>,
@@ -618,7 +618,7 @@ fn is_hidden(entry: &DirEntry) -> bool {
 #[cfg(test)]
 mod tests {
     use crate::id::ResourceId;
-    use crate::index::{discover_paths, scan_entries, IndexEntry};
+    use crate::index::{discover_paths, scan_entries, IndexEntry, IndexUpdate};
     use crate::initialize;
     use crate::ResourceIndex;
     use canonical_path::CanonicalPathBuf;
@@ -645,6 +645,8 @@ mod tests {
 
     const CRC32_1: u32 = 3817498742;
     const CRC32_2: u32 = 1804055020;
+
+    use crate::{ArklibError, Result};
 
     fn get_temp_dir() -> PathBuf {
         create_dir_at(std::env::temp_dir())
@@ -1151,6 +1153,48 @@ mod tests {
                 all_added.len().to_string(),
                 all_deleted.len().to_string()
             );
+        })
+    }
+
+    #[test]
+    fn should_correctly_update_all_compare_update_one_resource() {
+        run_test_and_clean_up(|path| {
+            create_file_at(path.clone(), Some(FILE_SIZE_2), Some(FILE_NAME_2));
+            let mut actual = ResourceIndex::build(path.clone());
+            create_file_at(path.clone(), Some(FILE_SIZE_1), Some(FILE_NAME_1));
+
+            let updated_using_update_all_result = actual.clone().update_all();
+
+            let mut all_deleted: HashSet<ResourceId> = HashSet::new();
+            let mut all_added: HashMap<CanonicalPathBuf, ResourceId> =
+                HashMap::new();
+
+            let curr_paths: HashMap<CanonicalPathBuf, DirEntry> =
+                discover_paths(path.clone());
+            let curr_entries = scan_entries(curr_paths);
+
+            for (cur_path, cur_entry) in curr_entries {
+                let one_index_update = actual
+                    .update_one(&cur_path.clone(), cur_entry.id)
+                    .expect("Should update one resource correctly");
+
+                for (add_path, &add_id) in one_index_update.added.iter() {
+                    all_added.insert(add_path.clone(), add_id.clone());
+                }
+
+                for delete_id in one_index_update.deleted {
+                    all_deleted.insert(delete_id);
+                }
+            }
+
+            let updated_using_update_one = IndexUpdate {
+                deleted: all_deleted,
+                added: all_added,
+            };
+            let updated_using_update_all =
+                updated_using_update_all_result.unwrap();
+
+            assert_eq!(updated_using_update_all, updated_using_update_one);
         })
     }
 }
--- a/src/index.rs
+++ b/src/index.rs
@@ -374,21 +374,23 @@ impl ResourceIndex {
                         self.forget_path(canonical_path, old_id)
                     }
                     Ok(new_entry) => {
+                        
                         // valid resource exists by the path
-                        let curr_entry = &self.path2id[canonical_path];
-
-                        if curr_entry.id == new_entry.id {
-                            // in rare cases we are here due to hash collision
-
-                            if curr_entry.modified == new_entry.modified {
-                                log::warn!("path {:?} was modified but not its content", &canonical_path);
+                        // in rare cases we are here due to hash collision
+                        if let Some(curr_entry) =
+                            self.path2id.get(canonical_path)
+                        {
+                            if curr_entry.id == new_entry.id {
+                                // in rare cases we are here due to hash collision
+                                if curr_entry.modified == new_entry.modified {
+                                    log::warn!("path {:?} was modified but not its content", &canonical_path);
+                                }
+                                // the caller must have ensured that the path was
+                                // updated
+                                return Err(ArklibError::Collision(
+                                    "New content has the same id".into(),
+                                ));
                             }
-
-                            // the caller must have ensured that the path was
-                            // updated
-                            return Err(ArklibError::Collision(
-                                "New content has the same id".into(),
-                            ));
                         }
 
                         // new resource exists by the path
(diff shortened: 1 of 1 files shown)


src/index.rs
```diff
@@ -374,21 +374,23 @@ impl ResourceIndex {
...
+                        
                         // valid resource exists by the path
-                        let curr_entry = &self.path2id[canonical_path];
-
-                        if curr_entry.id == new_entry.id {
-                            // in rare cases we are here due to hash collision
-
-                            if curr_entry.modified == new_entry.modified {
-                                log::warn!("path {:?} was modified but not its content", &canonical_path);
+                        // in rare cases we are here due to hash collision
```
kirillt, line 379: Incorrect comment (the real one is below).

kirillt:
> For instance, when the first parameter of the update_one method is the path of a newly added file, it returns an error.
> Furthermore, the same holds true when the first parameter of the update_one method corresponds to a file that has not been updated.

Interesting, thank you for this observation. Initially, the idea was that `update_one` must be called only on up-to-date index (when we know that exactly 1 path was updated). But it might be a good idea to allow updates on outdated index to make it more robust.

src/index.rs
```diff
@@ -374,21 +374,23 @@ impl ResourceIndex {
...
+                                // in rare cases we are here due to hash collision
+                                if curr_entry.modified == new_entry.modified {
+                                    log::warn!("path {:?} was modified but not its content", &canonical_path);
+                                }
+                                // the caller must have ensured that the path was
+                                // updated
+                                return Err(ArklibError::Collision(
+                                    "New content has the same id".into(),
+                                ));
                             }
```
kirillt, line 393: Might be worth to write something into the log in `else` branch.

funnyVOLT:
> > For instance, when the first parameter of the update_one method is the path of a newly added file, it returns an error.
> > Furthermore, the same holds true when the first parameter of the update_one method corresponds to a file that has not been updated.
> 
> Interesting, thank you for this observation. Initially, the idea was that `update_one` must be called only on up-to-date index (when we know that exactly 1 path was updated). But it might be a good idea to allow updates on outdated index to make it more robust.

Do you want to update the `update_one()`?

funnyVOLT:
> > For instance, when the first parameter of the update_one method is the path of a newly added file, it returns an error.
> > Furthermore, the same holds true when the first parameter of the update_one method corresponds to a file that has not been updated.
> 
> Interesting, thank you for this observation. Initially, the idea was that `update_one` must be called only on up-to-date index (when we know that exactly 1 path was updated). But it might be a good idea to allow updates on outdated index to make it more robust.

Would you like to update the update_one() method?

funnyVOLT:
> > For instance, when the first parameter of the update_one method is the path of a newly added file, it returns an error.
> > Furthermore, the same holds true when the first parameter of the update_one method corresponds to a file that has not been updated.
> 
> Interesting, thank you for this observation. Initially, the idea was that `update_one` must be called only on up-to-date index (when we know that exactly 1 path was updated). But it might be a good idea to allow updates on outdated index to make it more robust.

May I update 'update_one' method?

kirillt:
@funnyVOLT good stuff, but could you also compare results in the test? Sequence of `update_one` calls and `update_all` should provide exactly the same result. Would be great to have some assertions about it.

kirillt:
Something like this:
```
let mut index1 = index.clone();
index1.update_all(...);

let mut index2 = index;
for (path, id) in single_updates {
    index2.update_one(path, id);
}

assert_eq!(index1, index2);
```

funnyVOLT:
> Something like this:
> 
> ```
> let mut index1 = index.clone();
> index1.update_all(...);
> 
> let mut index2 = index;
> for (path, id) in single_updates {
>     index2.update_one(path, id);
> }
> 
> assert_eq!(index1, index2);
> ```
> 
> Cherry on top would be randomly generating `single_updates` and doing this assertion in a loop.

@kirillt, I already implemented similar test case in this function. 
https://github.com/funnyVOLT/arklib/blob/30cb80abb296d1a0b71345224545601af975d38a/src/index.rs#L1090

kirillt:
@funnyVOLT yes, I've seen this. But this test case doesn't compare output of `update_all` and `update_one`. It only asserts on output of `update_one`.

funnyVOLT:
> @funnyVOLT yes, I've seen this. But this test case doesn't compare output of `update_all` and `update_one`. It only asserts on output of `update_one`.

please let me check.
https://github.com/funnyVOLT/arklib/blob/a05e2d0300aef3824b4b347b443d2f60e466165c/src/index.rs#L1160

src/index.rs
```diff
@@ -352,6 +352,10 @@ impl ResourceIndex {
             return self.forget_id(old_id);
         }
 
+        let deleted: HashSet<ResourceId> = HashSet::new();
+        let added: HashMap<CanonicalPathBuf, ResourceId> = HashMap::new();
+        let mut update = IndexUpdate { deleted, added };
```
kirillt, line 357: These structures seem to be unused.

src/index.rs
```diff
@@ -365,33 +369,42 @@ impl ResourceIndex {
...
+                                log::warn!("Resourceid didn't modify but SystemTime was modified");
+                                // update.added.insert(
+                                //     canonical_path_buf.clone(),
+                                //     new_entry.id,
+                                // );
+                                // self.insert_entry(
+                                //     canonical_path_buf.clone(),
+                                //     new_entry,
+                                // );
+                                // update.deleted.insert(old_id);
```
kirillt, line 402: Commented-out code should be removed before merging.

src/index.rs
```diff
@@ -365,33 +369,42 @@ impl ResourceIndex {
...
+                                return Ok(update);
```
kirillt, line 403: It's better to keep returning `Err` as it was before. Not because it leads to incorrect state (it doesn't), but because the caller must ensure that the path being updated, was modified indeed. Otherwise, updating doesn't make sense and only spends computation resources.

src/index.rs
```diff
@@ -984,4 +1003,198 @@ mod tests {
         assert!(new2 > old2);
         assert!(new2 > new1);
     }
+
+    #[test]
+    fn should_correctly_update_one_resource() {
```
kirillt, line 1008: This test case is not correct. See test case `fn should_not_update_nonexistent_path`.

src/index.rs
funnyVOLT: > Commented-out code should be removed before merging.

@kirill I would like to ask you:
Why don't you do anything when the file has changed?

> Commented-out code should be removed before merging.

#kirill I would like to ask you: Why do you want to return "Err" when a file has been modified?
Shouldn't some information be returned when an existing file is modified?

src/index.rs
```diff
@@ -984,4 +1003,198 @@ mod tests {
...
+                update.deleted.len().to_string()
+            );
+
+            assert_eq!(update.added.len(), 1);
+            assert_eq!(actual.path2id.len(), 2);
+        })
+    }
+
+    #[test]
+    fn should_correctly_update_all_resource() {
```
kirillt, line 1054: This test case is a duplicate. See `fn should_update_resource_index_adding_1_additional_file_successfully`.

src/index.rs
```diff
@@ -984,4 +1003,198 @@ mod tests {
...
+            for (key, value) in &actual.path2id {
+                println!(
+                    "===❤❤========update_one_new_entry Resource ID: {}",
+                    value.id
+                );
+            }
+
+            create_file_at(path.clone(), Some(FILE_SIZE_1), Some(FILE_NAME_1));
+
+            for (old_path, old_index_enry) in &actual.path2id {
```
kirillt, line 1111: This is just 1 iteration, at this moment.

src/index.rs
```diff
@@ -984,4 +1003,198 @@ mod tests {
...
+                    .expect("Should update one resource correctly");
+
+                for (add_path, &add_id) in one_index_update.added.iter() {
+                    all_added.insert(add_path.clone(), add_id.clone());
+                }
+
+                for delete_id in one_index_update.deleted {
+                    all_deleted.insert(delete_id);
+                }
+            }
```
kirillt, line 1142: This test case contains no assertion, what is your expectation of the final state?

Also, please don't commit debugging prints (`println!`).

src/index.rs
```diff
@@ -984,4 +1003,198 @@ mod tests {
...
+
+            let updated_using_update_one = IndexUpdate {
+                deleted: all_deleted,
+                added: all_added,
+            };
```
kirillt, line 1193: It's a good idea to compare `IndexUpdate` structures, however we should focus first on `Index` instances themselves.

kirillt: Something like this is easier to start with:
```
    #[test]
    fn sequence_of_update_one_calls_should_imitate_update_all() {
        run_test_and_clean_up(|path| {
            create_file_at(path.clone(), Some(FILE_SIZE_2), Some(FILE_NAME_2));
            let index_initial = ResourceIndex::build(path.clone());
            create_file_at(path.clone(), Some(FILE_SIZE_1), Some(FILE_NAME_1));

            let mut index_full = index_initial.clone();
            index_full.update_all()
                .expect("Should update the index correctly");

            let curr_paths: HashMap<CanonicalPathBuf, DirEntry> =
                discover_paths(path.clone());
            let curr_entries = scan_entries(curr_paths);

            let mut index_by_one = index_initial.clone();

            // only 2 iterations at this moment
            for (cur_path, cur_entry) in curr_entries {
                index_by_one
                    .update_one(&cur_path.clone(), cur_entry.id)
                    .expect("Should update one resource correctly");
            }

            assert_eq!(index_full, index_by_one);
        })
    }
```

kirillt:
Superseded by #72

PR #72, opened 2024-01-20
#41: Add unit tests, fix update_one

kirillt:
Here are couple more tests.

`update_one` was handling absent paths incorrectly, fixed here.

Closing:
- #41

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -1,6 +1,5 @@
 use anyhow::anyhow;
 use canonical_path::{CanonicalPath, CanonicalPathBuf};
-use itertools::Itertools;
 use std::collections::{HashMap, HashSet};
 use std::fs::{self, File, Metadata};
 use std::io::{BufRead, BufReader, Write};
@@ -21,7 +20,7 @@ pub struct IndexEntry {
     pub id: ResourceId,
 }
 
-#[derive(Clone, Debug)]
+#[derive(PartialEq, Clone, Debug)]
 pub struct ResourceIndex {
     pub id2path: HashMap<ResourceId, CanonicalPathBuf>,
     pub path2id: HashMap<CanonicalPathBuf, IndexEntry>,
@@ -30,15 +29,26 @@ pub struct ResourceIndex {
     root: PathBuf,
 }
 
-#[derive(Debug)]
+#[derive(PartialEq, Debug, Default)]
 pub struct IndexUpdate {
     pub deleted: HashSet<ResourceId>,
     pub added: HashMap<CanonicalPathBuf, ResourceId>,
 }
 
+impl IndexUpdate {
+    pub fn added(mut self, path: CanonicalPathBuf, id: ResourceId) -> Self {
+        self.added.insert(path, id);
+        self
+    }
+    pub fn deleted(mut self, id: ResourceId) -> Self {
+        self.deleted.insert(id);
+        self
+    }
+}
+
 pub const RESOURCE_UPDATED_THRESHOLD: Duration = Duration::from_millis(1);
 
-type Paths = HashSet<CanonicalPathBuf>;
+pub type Paths = HashSet<CanonicalPathBuf>;
 
 impl ResourceIndex {
     pub fn size(&self) -> usize {
@@ -203,7 +213,7 @@ impl ResourceIndex {
 
         let curr_entries = discover_paths(self.root.clone());
 
-        //assuming that collections manipulation is
+        // assuming that collections manipulation is
         // quicker than asking `path.exists()` for every path
         let curr_paths: Paths = curr_entries.keys().cloned().collect();
         let prev_paths: Paths = self.path2id.keys().cloned().collect();
@@ -341,100 +351,128 @@ impl ResourceIndex {
         Ok(IndexUpdate { deleted, added })
     }
 
-    pub fn update_one(
+    /// The caller must ensure that:
+    /// * the index is up-to-date except this single path
+    /// * the path hasn't been indexed before
+    ///
+    /// Should only be used if reactive updates are not possible.
+    pub fn track_addition(
+        &mut self,
+        path: &dyn AsRef<Path>,
+    ) -> Result<IndexUpdate> {
+        log::debug!("Tracking a single addition in the index");
+
+        if !path.as_ref().exists() {
+            return Err(ArklibError::Path(
+                "Absent paths cannot be indexed".into(),
+            ));
+        }
+
+        let path_buf = CanonicalPathBuf::canonicalize(path)?;
+        let path = path_buf.as_canonical_path();
+
+        let metadata = fs::metadata(path).map_err(|_| {
+            ArklibError::Path("Couldn't retrieve file metadata".into())
+        })?;
+        let entry = scan_entry(path, metadata).map_err(|_| {
+            ArklibError::Path(
+                "The path points to a directory or empty file".into(),
+            )
+        })?;
+        let result = IndexUpdate::default().added(path_buf.clone(), entry.id);
+        self.insert_entry(path_buf, entry);
+        Ok(result)
+    }
+
+    /// The caller must ensure that:
+    /// * the index is up-to-date except this single id
+    /// * the resource with this id has been indexed before
+    /// * the resource with this id doesn't exist anymore
+    ///
+    /// Should only be used if reactive updates are not possible.
+    pub fn track_deletion(&mut self, id: ResourceId) -> Result<IndexUpdate> {
+        log::debug!("Tracking a single deletion in the index");
+
+        let indexed_path = self.id2path.get(&id);
+        if indexed_path.is_none() {
+            return Err(ArklibError::Path(
+                "The id cannot be found in the index".into(),
+            ));
+        }
+
+        let indexed_path = indexed_path.unwrap().clone();
+        self.forget_entry(indexed_path.as_canonical_path(), id)?;
+
+        Ok(IndexUpdate::default().deleted(id))
+    }
+
+    /// The caller must ensure that:
+    /// * the index is up-to-date except this single path
+    /// * the path has been indexed before
+    /// * the path has been mapped into `old_id`
+    ///
+    /// Should only be used if reactive updates are not possible.
+    /// Returns an empty update if the resource hasn't been really updated.
+    pub fn track_update(
         &mut self,
         path: &dyn AsRef<Path>,
         old_id: ResourceId,
     ) -> Result<IndexUpdate> {
-        log::debug!("Updating a single entry in the index");
+        log::debug!("Tracking a single update in the index");
 
-        if !path.as_ref().exists() {
-            return self.forget_id(old_id);
-        }
-
-        let canonical_path_buf = CanonicalPathBuf::canonicalize(path)?;
-        let canonical_path = canonical_path_buf.as_canonical_path();
-
-        log::trace!(
-            "[update] paths {:?} has id {:?}",
-            canonical_path,
-            self.path2id[canonical_path]
-        );
-
-        match fs::metadata(canonical_path) {
-            Err(_) => {
-                // updating the index after resource removal is a correct
-                // scenario
-                self.forget_path(canonical_path, old_id)
-            }
-            Ok(metadata) => {
-                match scan_entry(canonical_path, metadata) {
-                    Err(_) => {
-                        // a directory or empty file exists by the path
-                        self.forget_path(canonical_path, old_id)
-                    }
-                    Ok(new_entry) => {
-                        // valid resource exists by the path
-                        let curr_entry = &self.path2id[canonical_path];
-
-                        if curr_entry.id == new_entry.id {
-                            // in rare cases we are here due to hash collision
-
-                            if curr_entry.modified == new_entry.modified {
-                                log::warn!("path {:?} was modified but not its content", &canonical_path);
-                            }
-
-                            // the caller must have ensured that the path was
-                            // updated
-                            return Err(ArklibError::Collision(
-                                "New content has the same id".into(),
-                            ));
-                        }
-
-                        // new resource exists by the path
-                        self.forget_path(canonical_path, old_id).map(
-                            |mut update| {
-                                update.added.insert(
-                                    canonical_path_buf.clone(),
-                                    new_entry.id,
-                                );
-                                self.insert_entry(
-                                    canonical_path_buf,
-                                    new_entry,
-                                );
-
-                                update
-                            },
-                        )
-                    }
-                }
+        if let Some(indexed_path) = self.id2path.get(&old_id) {
+            if indexed_path.as_path() != path.as_ref() {
+                return Err(ArklibError::Path(
+                    "The path isn't indexed or doesn't map to the id".into(),
+                ));
             }
         }
-    }
 
-    pub fn forget_id(&mut self, old_id: ResourceId) -> Result<IndexUpdate> {
-        let old_path = self
-            .path2id
-            .drain()
-            .filter_map(|(k, v)| {
-                if v.id == old_id {
-                    Some(k)
+        return fs::metadata(path)
+            .map_err(ArklibError::Io)
+            .and_then(|metadata| {
+                let path_buf = CanonicalPathBuf::canonicalize(path)?;
+                let path = path_buf.as_canonical_path();
+
+                log::trace!(
+                    "[update] paths {:?} has id {:?}",
+                    path,
+                    self.path2id[path]
+                );
+
+                let curr_entry = match self.path2id.get(path) {
+                    Some(entry) => entry,
+                    None => {
+                        return Err(ArklibError::Path(
+                            "The path hasn't been indexed before".into(),
+                        ))
+                    }
+                };
+
+                let new_entry = scan_entry(path, metadata)?;
+
+                if curr_entry.id != new_entry.id {
+                    let mut deleted = HashSet::new();
+                    deleted.insert(old_id);
+
+                    let mut added = HashMap::new();
+                    added.insert(path_buf.clone(), new_entry.id);
+
+                    self.forget_entry(path, old_id)?;
+                    self.insert_entry(path_buf, new_entry);
+
+                    Ok(IndexUpdate { deleted, added })
+                } else if curr_entry.modified == new_entry.modified {
+                    log::warn!("path {:?} was not modified", &path);
+                    Ok(IndexUpdate::default())
                 } else {
-                    None
+                    log::warn!(
+                        "path {:?} was modified but not its content",
+                        &path
+                    );
+                    Ok(IndexUpdate::default())
                 }
-            })
-            .collect_vec();
-        for p in old_path {
-            self.path2id.remove(&p);
-        }
-        self.id2path.remove(&old_id);
-        let mut deleted = HashSet::new();
-        deleted.insert(old_id);
-
-        Ok(IndexUpdate {
-            added: HashMap::new(),
-            deleted,
-        })
+            });
     }
 
     fn insert_entry(&mut self, path: CanonicalPathBuf, entry: IndexEntry) {
@@ -454,61 +492,57 @@ impl ResourceIndex {
         self.path2id.insert(path, entry);
     }
 
-    fn forget_path(
+    fn forget_entry(
         &mut self,
         path: &CanonicalPath,
-        old_id: ResourceId,
-    ) -> Result<IndexUpdate> {
-        self.path2id.remove(path);
-
-        if let Some(collisions) = self.collisions.get_mut(&old_id) {
+        id: ResourceId,
+    ) -> Result<()> {
+        let removed = self.path2id.remove(path);
+        let removed = match removed {
+            Some(entry) => entry,
+            None => {
+                return Err(ArklibError::Path("The path is not indexed".into()))
+            }
+        };
+        debug_assert_eq!(removed.id, id, "The path is mapped to another id");
+        if let Some(collisions) = self.collisions.get_mut(&id) {
             debug_assert!(
                 *collisions > 1,
                 "Any collision must involve at least 2 resources"
             );
             *collisions -= 1;
-
             if *collisions == 1 {
-                self.collisions.remove(&old_id);
+                self.collisions.remove(&id);
             }
-
             // minor performance issue:
             // we must find path of one of the collided
             // resources and use it as new value
-            let maybe_collided_path =
+            let collided_path =
                 self.path2id.iter().find_map(|(path, entry)| {
-                    if entry.id == old_id {
+                    if entry.id == id {
                         Some(path)
                     } else {
                         None
                     }
                 });
-
-            if let Some(collided_path) = maybe_collided_path {
-                let old_path =
-                    self.id2path.insert(old_id, collided_path.clone());
-
-                debug_assert_eq!(
-                    old_path.unwrap().as_canonical_path(),
-                    path,
-                    "Must forget the requested path"
-                );
-            } else {
-                return Err(ArklibError::Collision(
-                    "Illegal state of collision tracker".into(),
-                ));
-            }
+            let collided_path = match collided_path {
+                Some(path) => path.clone(),
+                None => {
+                    return Err(ArklibError::Other(anyhow!(
+                        "Illegal state of collision tracker"
+                    )))
+                }
+            };
+            let old_path = self.id2path.insert(id, collided_path.clone());
+            debug_assert_eq!(
+                old_path.unwrap().as_canonical_path(),
+                path,
+                "Must forget the requested path"
+            );
         } else {
-            self.id2path.remove(&old_id);
+            self.id2path.remove(&id);
         }
-
-        let mut deleted = HashSet::new();
-        deleted.insert(old_id);
-
-        Ok(IndexUpdate {
-            added: HashMap::new(),
-            deleted,
-        })
+        Ok(())
     }
 }
 
@@ -557,10 +591,7 @@ fn scan_entry(path: &CanonicalPath, metadata: Metadata) -> Result<IndexEntry> {
 
     let size = metadata.len();
     if size == 0 {
-        Err(std::io::Error::new(
-            std::io::ErrorKind::InvalidData,
-            "Empty resource",
-        ))?;
+        return Err(ArklibError::Path("Resource cannot be empty".into()));
     }
 
     let id = ResourceId::compute(size, path)?;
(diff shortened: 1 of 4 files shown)


src/index.rs
```diff
@@ -403,7 +403,13 @@ impl ResourceIndex {
         log::debug!("Updating a single entry in the index");
 
         if !path.as_ref().exists() {
-            return self.forget_id(old_id);
+            if let Some(indexed_path) = self.id2path.get(&old_id) {
+                if indexed_path.as_path() == path { //FIXME
+                    return self.forget_id(old_id);
+                }
```
kirillt, line 409: My Rust skills are not so strong, how to compare paths here? The problem is that `path` is generic...

We need this piece for correct handling absent paths.

kirillt:
- [ ] correct comparison of generic path against fixed-typed path
- [ ] implement automatic test by this idea
    https://github.com/ARK-Builders/arklib/pull/71#discussion_r1460340333

src/index.rs
funnyVOLT: > My Rust skills are not so strong, how to compare paths here? The problem is that `path` is generic...
> 
> We need this piece for correct handling absent paths.

@kirill. using this code.
if indexed_path.as_path() == path.as_ref() { 
                    let path_buf: PathBuf = indexed_path.into();
                    println!("==indexed_path= {}==path={}==",path_buf.to_str().unwrap_or("no_path_buf"), path.as_ref().to_str().unwrap_or("no_path"));
                    return self.forget_id(old_id);
                }

kirillt: @funnyVOLT thank you!

src/index.rs
```diff
@@ -403,7 +403,15 @@ impl ResourceIndex {
...
-            return self.forget_id(old_id);
+            if let Some(indexed_path) = self.id2path.get(&old_id) {
+                if indexed_path.as_path() == path.as_ref() {
+                    return self.forget_id(old_id);
+                }
+            }
+
+            return Err(ArklibError::Path(
+                "The path isn't indexed or doesn't map to the id".into(),
+            ));
```
kirillt, line 414: We throw this error if the caller provided wrong `old_id`. Alternative would be querying `path2id` by `path`, but this would not cover the case with resources that were indexed but then disappeared from the filesystem.

Alternatively, we can extract such case into `unindex_one`. The caller would need to be more precise about what to call when, but it's always possible. The app that calls the `update_one` or `index_new` function must always have information, otherwise complete re-indexing with `update_all` must be performed.

Usage scenarios:
1. User scrolls through images in Gallery-like app, presses `delete` button. We don't want to run full re-indexing because it's slow. With reactive updates the index must be up-to-date anyway. We precisely know that only this single resource requires update and we call `update_one` (or `unindex_one`).
2. User scrolls through images in Gallery again, now they press `edit` button. We pass the image to another app but we have callback when we come back to Gallery. In such a fallback, `update_one` must be called. The app always has old id and path.
3. User opens graphical editor, creates an image from scratch, selects ARK-enabled folder, so we have an index in there by path `.ark/index`. Function `index_new` can be called to add single path to the index. However, I'm not entirely sure this use-case is useful. Because, other apps either monitor the folder at the same moment and catch update automatically, or they don't monitor and will perform full re-indexing anyway. So `index_new` might be useless.

kirillt: 3'. User uses Navigator to copy a batch of files from one root to another root. Navigator can use `unindex_one` and `index_new` to update the index in both source and destination folders. However, again, with reactive updates this should be done automatically. Is it worth adding new functions?

funnyVOLT:
@kirill.
I would like to inform you of my opinion.
The problem occurred while comparing 'update_all' and 'track_methods'.
1. When a new file is created.
  When files with the same ResourceID are added, update_all and track_addition operate differently.
2. When a new file is modified
  After the file is changed, update_all does not seem to reflect the changed information.
  The handling of collisions in the 'track_update' function is not reflected clearly.
3. Overall, I think I need to check 'collisions' again.

funnyVOLT:
@kirill.
I would like to inform you of my opinion.
The problem occurred while comparing 'update_all' and 'track_methods'.
1. When a new file is created.
  When files with the same ResourceID are added, update_all and track_addition operate differently.
2. When a new file is modified
  After the file is changed, update_all does not seem to reflect the changed information.
  The handling of collisions in the 'track_update' function is not reflected clearly.
3. Overall, I think I need to check 'collisions' again.

kirillt:
@funnyVOLT thank you for the test case. I'll look into it deeper soon. It's great that you've discovered corner cases I've missed!

Could you create small isolated test cases reproducing these 2 problems?
> 1. When files with the same ResourceID are added, `update_all` and `track_addition` operate differently.

>  2. After the file is changed, `update_all` does not seem to reflect the changed information.

funnyVOLT:
> @funnyVOLT , PLEASE USE @kirillt , not me. Thank you very much for spam

thank you.

src/index.rs
```diff
@@ -1133,4 +1132,211 @@ mod tests {
...
+        folder_2.push(FILE_DIR_2);
+
+        std::fs::create_dir(&folder_1).expect("Could not create temp dir");
+        std::fs::create_dir(&folder_2).expect("Could not create temp dir");
+
+        for i in 1..FILE_COUNT {
+            let data_size: u64 = (i + 10).try_into().unwrap();
+            let mut create_file_name = String::from(FILE_NAME);
+            create_file_name.push_str(&i.to_string());
+            create_file_name.push_str(".txt");
```
kirillt, line 1209: File extension is not necessary.

src/index.rs
```diff
@@ -1133,4 +1132,211 @@ mod tests {
...
+            // Initial folders and files are created in this function.
+            init_tmp_directory(path.clone());
+            let initial_index = ResourceIndex::build(path.clone());
+            let mut index1 = initial_index.clone();
+            let mut index2 = initial_index.clone();
+
+            for i in 1..FILE_COUNT {
+                let mut file_name = String::from("test_");
+                file_name.push_str(&i.to_string());
+                file_name.push_str(".txt");
```
kirillt, line 1231: File extension isn't necessary.

src/index.rs
```diff
@@ -1133,4 +1132,211 @@ mod tests {
...
+        let mut folder_2 = root_path.clone();
+        folder_1.push(FILE_DIR_1);
+        folder_2.push(FILE_DIR_2);
+
+        std::fs::create_dir(&folder_1).expect("Could not create temp dir");
+        std::fs::create_dir(&folder_2).expect("Could not create temp dir");
+
+        for i in 1..FILE_COUNT {
+            let data_size: u64 = (i + 10).try_into().unwrap();
+            let mut create_file_name = String::from(FILE_NAME);
```
kirillt, line 1207: Let's create a function returning random unique filename.

src/index.rs
```diff
@@ -1133,4 +1132,211 @@ mod tests {
...
+        if let Some(file_name) = name {
+            cur_file_name = file_name.clone();
+            cur_file_path.push(file_name);
+        }
+
+        match rnd_num {
+            // create
+            1 => {
+                let mut file_name_new = String::from(cur_file_name);
+                file_name_new.push_str("_new.txt");
```
kirillt, line 1165: Would be better to provide new random unique name, automatically.

src/index.rs
```diff
@@ -1133,4 +1132,211 @@ mod tests {
...
+                create_file_at(
+                    folder_1.clone(),
+                    Some(DATA_SIZE_2),
+                    Some(&file_name_new),
+                );
+            }
+            // update
+            2 => {
+                let mut file = File::create(cur_file_path.as_path())
```
kirillt, line 1174: The file must exist already, right?

src/index.rs
```diff
@@ -1133,4 +1132,211 @@ mod tests {
...
+        folder_name: Option<&str>,
+        name: Option<&str>,
+    ) -> i32 {
+        let mut rng = rand::thread_rng();
+        let mut rnd_num = rng.gen_range(1..=4);
+
+        let mut folder_1 = root_path.clone();
+        let mut folder_2 = root_path.clone();
+        folder_1.push(FILE_DIR_1);
+        folder_2.push(FILE_DIR_2);
```
kirillt, line 1152: Do we really need 2 folders?

src/index.rs
funnyVOLT: > File extension is not necessary.

I deleted line 1209.

src/index.rs
funnyVOLT: > File extension isn't necessary.

I deleted line 1231.

src/index.rs
funnyVOLT: > Would be better to provide new random unique name, automatically.

I added const variable CREATE, UPDATE, DELETE, MOVE .

src/index.rs
funnyVOLT: > Do we really need 2 folders?

It is necessary for file movement function.

src/index.rs
funnyVOLT: > The file must exist already, right?

yes.
But, In tis unit test, the file always exists.

src/index.rs
funnyVOLT: > Would be better to provide new random unique name, automatically.
I created a random unique name.

src/index.rs
funnyVOLT: > Let's create a function returning random unique filename.

I created a random unique name.

src/index.rs
```diff
@@ -1133,18 +1133,19 @@ mod tests {
         assert!(new2 > new1);
     }
 
+    const CREATE: i32 = 1;
+    const UPDATE: i32 = 2;
+    const DELETE: i32 = 3;
+    const MOVE: i32 = 5;
```
kirillt, line 1139: Why `MOVE` is `5`?

Actually, it's better to use `enum` to avoid mistakes with numbers.

src/index.rs
```diff
@@ -1133,18 +1133,19 @@ mod tests {
...
 
-    fn generate_random_update(
-        root_path: PathBuf,
-        folder_name: Option<&str>,
-        name: Option<&str>,
-    ) -> i32 {
+    fn generate_random_update(root_path: PathBuf, name: Option<&str>) -> i32 {
         let mut rng = rand::thread_rng();
-        let mut rnd_num = rng.gen_range(1..=4);
+        let rand_num: i32 = rng.gen_range(0..10);
```
kirillt, line 1148: What is `10`? Better to declare the constant.

src/index.rs
```diff
@@ -1154,34 +1155,36 @@ mod tests {
...
-            1 => {
-                let mut file_name_new = String::from(cur_file_name);
-                file_name_new.push_str("_new.txt");
+        match rand_num {
+            CREATE => {
+
+                // create randomize file name
+                let random_file_name_new: String = rng
+                .sample_iter(&Alphanumeric)
+                .take(7)
```
kirillt, line 1168: Let's declare `7` as a named constant.

src/index.rs
```diff
@@ -1336,7 +1333,7 @@ mod tests {
                 .update_all()
                 .expect("Should update index correctly");
 
-            assert_eq!(initial_index , index_update_all);
+            assert!(initial_index != index_update_all);
```
kirillt, line 1336: `assert_ne!`

kirillt, line 1336: `assert_ne!`

src/index.rs
kirillt: I see only constant `FILE_NAME`.

src/index.rs
funnyVOLT: updated like this code.
assert_ne!(
                initial_index, index_update_all,
                "we are testing that the values are not equal"
            );

src/index.rs
funnyVOLT: > Let's declare `7` as a named constant.

const FILE_RAND_NAME_SIZE: usize = 7;

src/index.rs
funnyVOLT: > What is `10`? Better to declare the constant.

I update this code.
let rand_num: i32 = rng.gen_range(0..ACTIONCNT);

src/index.rs
kirillt: We can also just move to new random filename. This way we can apply move operation multiple times, no need to track in which of the two the file resides.

src/index.rs
```diff
@@ -1156,56 +1155,60 @@ mod tests {
         }
     }
 
-    const FILE_DIR_1: &str = "folder_1";
-    const FILE_DIR_2: &str = "folder_2";
     const FILE_NAME: &str = "test_";
-    const FILE_COUNT: i32 = 10;
+    const FILE_COUNT: i32 = 3;
```
kirillt, line 1159: Isn't 3 too few for proper testing?

src/index.rs
```diff
@@ -1320,7 +1319,7 @@ mod tests {
                 .update_all()
                 .expect("Should update index correctly");
 
-            assert_eq!(index_track_addition, index_update_all);
+            assert_ne!(index_track_addition, index_update_all);
```
kirillt: Must be equal, isn't it?

src/index.rs
```diff
@@ -1133,4 +1133,225 @@ mod tests {
...
+            let mut added_file_path = path.clone();
+            added_file_path.push(FILE_NAME_2);
+            let _ = index_track_addition.track_addition(&added_file_path);
+
+            index_update_all
+                .update_all()
+                .expect("Should update index correctly");
+
+            assert_ne!(
+                index_track_addition, index_update_all,
```
kirillt, line 1322: Must be equal anyway, isn't it?

src/index.rs
```diff
@@ -1133,4 +1133,225 @@ mod tests {
...
+
+            let mut added_file_path = path.clone();
+            added_file_path.push(FILE_NAME_2);
+            let _ = index_track_addition.track_addition(&added_file_path);
+
+            index_update_all
+                .update_all()
+                .expect("Should update index correctly");
+
+            assert_ne!(
```
kirillt, line 1321: After this assertion we should `sleep` for 1ms, perform `update_all` again and `assert_eq` new index.

src/index.rs
```diff
@@ -36,6 +35,35 @@ pub struct IndexUpdate {
     pub added: HashMap<CanonicalPathBuf, ResourceId>,
 }
 
+impl IndexUpdate {
+    pub fn empty() -> Self {
+        IndexUpdate {
+            deleted: HashSet::new(),
+            added: HashMap::new(),
+        }
+    }
```
tareknaser, line 44: We can implement `Default` instead

```rust
impl Default for IndexUpdate {
    fn default() -> Self {
        IndexUpdate {
            deleted: HashSet::new(),
            added: HashMap::new(),
        }
    }
}
```

tareknaser, line 44: We can implement Default instead
```rust
impl Default for IndexUpdate {  
    fn default() -> Self {  
        IndexUpdate {  
            deleted: HashSet::new(),  
            added: HashMap::new(),  
        }  
    }  
} 
```

src/index.rs
```diff
@@ -36,6 +35,35 @@ pub struct IndexUpdate {
...
+
+    pub fn deleted(id: ResourceId) -> Self {
+        let mut deleted = HashSet::new();
+        deleted.insert(id);
+
+        IndexUpdate {
+            deleted,
+            added: HashMap::new(),
+        }
+    }
```
tareknaser, line 64:
```suggestion
    pub fn added(mut self, path: CanonicalPathBuf, id: ResourceId) -> Self {
        self.added.insert(path, id);
        self
    }

    pub fn deleted(mut self, id: ResourceId) -> Self {
        self.deleted.insert(id);
        self
    }
```

and then we could use `IndexUpdate::default().added(path_buf, entry.id);`

src/index.rs
```diff
@@ -356,142 +389,123 @@ impl ResourceIndex {
...
-                    return Err(ArklibError::Path(
-                        "The path points to a directory or empty file".into(),
-                    ));
+                Err(_) => Err(ArklibError::Path(
+                    "The path points to a directory or empty file".into(),
+                )),
+                Ok(entry) => {
+                    let result = IndexUpdate::added(path_buf.clone(), entry.id);
+                    self.insert_entry(path_buf, entry);
+                    Ok(result)
```
tareknaser, line 403: We can also do this to reduce the indentation levels and improve readability.

```suggestion
        let metadata = fs::metadata(path).map_err(|_| {
            ArklibError::Path("Couldn't retrieve file metadata".into())
        })?;

        let entry = scan_entry(path, metadata).map_err(|_| {
            ArklibError::Path(
                "The path points to a directory or empty file".into(),
            )
        })?;

        let result = IndexUpdate::added(path_buf.clone(), entry.id);
        self.insert_entry(path_buf, entry);
        Ok(result)
```

src/index.rs
```diff
@@ -341,11 +369,16 @@ impl ResourceIndex {
         Ok(IndexUpdate { deleted, added })
     }
 
-    // the caller must ensure that:
+    // The caller must ensure that:
     // * the index is up-to-date except this single path
     // * the path hasn't been indexed before
```
tareknaser, line 374: Let's make these [doc comments](https://doc.rust-lang.org/reference/comments.html#doc-comments)
just change "//" to "///" before all functions

src/index.rs
```diff
@@ -356,142 +389,123 @@ impl ResourceIndex {
...
+                    path,
+                    self.path2id[path]
+                );
+
+                let curr_entry = &self.path2id.get(path);
+                if curr_entry.is_none() {
+                    return Err(ArklibError::Path("The path hasn't been indexed before".into()));
+                }
+
+                let curr_entry = curr_entry.unwrap();
```
tareknaser, line 470:
```suggestion
                let curr_entry = match self.path2id.get(path) {
                    Some(entry) => entry,
                    None => return Err(ArklibError::Path("The path hasn't been indexed before".into())),
                };
```

src/index.rs
```diff
@@ -356,142 +389,123 @@ impl ResourceIndex {
...
-                        self.forget_path(path, old_id).map(|mut update| {
-                            update
-                                .added
-                                .insert(path_buf.clone(), new_entry.id);
-                            self.insert_entry(path_buf, new_entry);
-
-                            update
-                        })
-                    }
+                    }),
```
tareknaser, line 506: we can also do
```rust
                let new_entry = scan_entry(path, metadata)?;

                if curr_entry.id != new_entry.id {
                    let mut deleted = HashSet::new();
                    deleted.insert(old_id);

                    let mut added = HashMap::new();
                    added.insert(path_buf.clone(), new_entry.id);

                    self.forget_entry(path, old_id);
                    self.insert_entry(path_buf, new_entry);

                    Ok(IndexUpdate { deleted, added })
                } else if curr_entry.modified == new_entry.modified {
                    log::warn!("path {:?} was not modified", &path);
                    Ok(IndexUpdate::empty())
                } else {
                    log::warn!(
                        "path {:?} was modified but not its content",
                        &path
                    );
                    Ok(IndexUpdate::empty())
                }
```

src/index.rs
```diff
@@ -511,61 +525,51 @@ impl ResourceIndex {
...
-
-        let mut deleted = HashSet::new();
-        deleted.insert(old_id);
-
-        Ok(IndexUpdate {
-            added: HashMap::new(),
-            deleted,
-        })
     }
 }
```
tareknaser, line 574: We should avoid using `unwrap()`. Here is better error handling

```suggestion
    fn forget_entry(
        &mut self,
        path: &CanonicalPath,
        id: ResourceId,
    ) -> Result<()> {
        let removed = self.path2id.remove(path);
        let removed = match removed {
            Some(entry) => entry,
            None => {
                return Err(ArklibError::Path("The path is not indexed".into()))
            }
        };

        debug_assert_eq!(removed.id, id, "The path is mapped to another id");

        if let Some(collisions) = self.collisions.get_mut(&id) {
            debug_assert!(
                *collisions > 1,
                "Any collision must involve at least 2 resources"
            );
            *collisions -= 1;

            if *collisions == 1 {
                self.collisions.remove(&id);
            }

            // minor performance issue:
            // we must find path of one of the collided
            // resources and use it as new value
            let collided_path =
                self.path2id.iter().find_map(|(path, entry)| {
                    if entry.id == id {
                        Some(path)
                    } else {
                        None
                    }
                });

            let collided_path = match collided_path {
                Some(path) => path.clone(),
                None => {
                    return Err(ArklibError::Other(anyhow!(
                        "Illegal state of collision tracker"
                    )))
                }
            };

            let old_path = self.id2path.insert(id, collided_path.clone());
            debug_assert_eq!(
                old_path.unwrap().as_canonical_path(),
                path,
                "Must forget the requested path"
            );
        } else {
            self.id2path.remove(&id);
        }
        Ok(())
    }
```

src/index.rs
```diff
@@ -888,191 +979,100 @@ mod tests {
...
 
     #[test]
-    fn update_all_should_error_on_files_without_permissions() {
-        run_test_and_clean_up(|path| {
-            create_file_at(path.clone(), Some(FILE_SIZE_1), Some(FILE_NAME_1));
-            let (file, _) = create_file_at(
-                path.clone(),
-                Some(FILE_SIZE_2),
-                Some(FILE_NAME_2),
+    fn track_update_should_index_modified_paths() {
```
tareknaser, line 1019: This test fails on my mac 

```bash
---- index::tests::track_update_should_index_modified_paths stdout ----
thread 'index::tests::track_update_should_index_modified_paths' panicked at src/index.rs:1032:18:
called `Result::unwrap()` on an `Err` value: Path("The path isn't indexed or doesn't map to the id")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'index::tests::track_update_should_index_modified_paths' panicked at src/index.rs:758:13:
Test panicked
```

tareknaser (changes requested):
Great job!
Left some comments.

I think we can also add simple doc comments on the methods implemented for `ResourceIndex`

src/index.rs
kirillt: Oh yeah, that would be much better

src/index.rs
kirillt: This looks really great!

src/index.rs
tareknaser: regarding this test,
Why should the result from `track_addition` and `update_all` be different?

src/index.rs
```diff
@@ -715,6 +698,26 @@ mod tests {
...
+            data_size: DATA_SIZE_1,
+            crc32: CRC32_1,
+        }
+    }
+
+    fn resource_id_2() -> ResourceId {
+        ResourceId {
+            data_size: DATA_SIZE_2,
+            crc32: CRC32_2,
+        }
```
tareknaser, line 718: Let's define these resources as constants instead

```suggestion
   const RESOURCE_ID_1: ResourceId = ResourceId {
        data_size: DATA_SIZE_1,
        crc32: CRC32_1,
    };

    const RESOURCE_ID_2: ResourceId = ResourceId {
        data_size: DATA_SIZE_2,
        crc32: CRC32_2,
    };
```

src/index.rs
```diff
@@ -669,15 +643,24 @@ mod tests {
     use std::fs::File;
     #[cfg(target_os = "linux")]
     use std::fs::Permissions;
+    use std::io::Write;
     #[cfg(target_os = "linux")]
     use std::os::unix::fs::PermissionsExt;
```
tareknaser, line 648: This gives us permissions to write files on linux systems only so some unit tests are failing on mac.

We can use `#[cfg(target_family = "unix")]` instead

src/index.rs
```diff
@@ -715,6 +698,26 @@ mod tests {
...
+    }
+
+    fn resource_id_2() -> ResourceId {
+        ResourceId {
+            data_size: DATA_SIZE_2,
+            crc32: CRC32_2,
+        }
+    }
+
     fn run_test_and_clean_up(
```
tareknaser, line 721: I am not sure we need this function
we can use `TempDir::new` to achieve the same output

[`tempdir` crate automatically cleans up the temporary directories when they go out of scope](https://docs.rs/tempdir/latest/tempdir/struct.TempDir.html)

src/index.rs
```diff
@@ -1133,4 +1106,226 @@ mod tests {
...
+                    }
+                    Ok(FileAction::DELETE) => {
+                        let _ = index1.track_deletion(old_id);
+                    }
+                    Ok(FileAction::MOVE) => {
+                        let _ = index1.track_deletion(old_id);
+                        let _ = index1.track_addition(&new_file_path);
+                    }
+                    Err(_) => println!("rnd_num error"),
+                }
```
tareknaser, line 1266: Are we intentionally ignoring errors here for testing purposes?


If not, we should panic with errors

```suggestion
               match update_state.try_into() {
                    Ok(FileAction::CREATE) => {
                        index1
                            .track_addition(&new_file_path)
                            .expect("Should track addition correctly");
                    }
                    Ok(FileAction::UPDATE) => {
                        index1
                            .track_update(&file_path, old_id)
                            .expect("Should track update correctly");
                    }
                    Ok(FileAction::DELETE) => {
                        index1
                            .track_deletion(old_id)
                            .expect("Should track deletion correctly");
                    }
                    Ok(FileAction::MOVE) => {
                        index1
                            .track_deletion(old_id)
                            .expect("Should track deletion correctly");
                        index1
                            .track_addition(&new_file_path)
                            .expect("Should track addition correctly");
                    }
                    Err(_) => println!("rnd_num error"),
                }
```

src/index.rs
```diff
@@ -1133,4 +1106,226 @@ mod tests {
...
+
+            let initial_index = ResourceIndex::build(path.clone());
+            let mut index_track_addition = initial_index.clone();
+            let mut index_update_all = initial_index.clone();
+
+            create_file_at(path.clone(), Some(DATA_SIZE_1), Some(FILE_NAME_2));
+
+            let mut added_file_path = path.clone();
+            added_file_path.push(FILE_NAME_2);
+            let _ = index_track_addition.track_addition(&added_file_path);
```
tareknaser, line 1288: Here as well

src/index.rs
```diff
@@ -1133,4 +1106,226 @@ mod tests {
...
+                std::fs::rename(cur_file_path, new_file_path)
+                    .expect("Should rename file successfully");
+            }
+            Err(_) => println!("rnd_num error"),
+        }
+
+        return rand_num;
+    }
+
+    fn init_tmp_directory(root_path: PathBuf) {
```
tareknaser, line 1195: This function deserves a better name.

Maybe `create_temp_files_in_directory`

src/index.rs
```diff
@@ -1133,4 +1106,226 @@ mod tests {
...
+        fn from(value: i32) -> Self {
+            match value {
+                0 => FileAction::CREATE,
+                1 => FileAction::UPDATE,
+                2 => FileAction::DELETE,
+                3 => FileAction::MOVE,
+                _ => unreachable!("Invalid ActionType value"),
+            }
+        }
+    }
```
tareknaser, line 1129: Maybe it's better to keep all the constants and functions that are not tests at the top of the testing module.

tareknaser (commented):
some more comments but we are getting there

src/index.rs
tareknaser: For context: https://github.com/ARK-Builders/arklib/pull/72#discussion_r1478470500

src/index.rs
tareknaser: I will leave this one for a follow up PR

src/index.rs
kirillt: @tareknaser yes, we must panic and not ignore the errors. But maybe simple unwrapping would be fine in test code?

kirillt:
The scope of this PR was split and reworked in new repo:
- https://github.com/ARK-Builders/ark-core/pull/36
- https://github.com/ARK-Builders/ark-core/pull/79
- https://github.com/ARK-Builders/ark-core/pull/85

Issue #65, opened 2024-01-05
Benchmark `crc32` and `blake3` hash functions


kirillt:
We use [crc32fast](https://github.com/srijs/rust-crc32fast) crate to generate `ResourceId`. It was one of the fastest hash functions 3 years ago, when [blake3](https://github.com/BLAKE3-team/BLAKE3) was invented.

Official metrics:
* Blake3, _AWS c5.metal, 16 KiB input, 1 thread_:
**6866** MiB/s

* CRC32, _unknown env and run parameters_:
baseline: 1499 MiB/s,
pclmulqdq: **7314** MiB/s

Let's create a small benchmark to compare them in same environment, with same parameters.

Even if `blake3` is same performant as `crc32`, it would be worth updating `arklib`, because `blake3` is cryptographic hash function and it means no collisions in the index.

kirillt:
Different parameters and different cores number should be tested in the benchmark.

kirillt:
This [benchmark](https://github.com/rurban/smhasher) presents different results:
> Hash function 	**MiB/sec** 	_cycl./hash 	cycl./map 	size_ 	Quality problems
> [blake3_c](https://github.com/rurban/smhasher/blob/master/doc/blake3_c.txt) 	**1288.84** 	_357.69_ 	_582.89 (6)_ 		no 32bit portability
> [crc32](https://github.com/rurban/smhasher/blob/master/doc/crc32.txt) 	**383.12** 	_134.21 	257.50 (11) 	422_ 	insecure, 8590x collisions, distrib, PerlinNoise
> [crc32_hw](https://github.com/rurban/smhasher/blob/master/doc/crc32_hw.txt) 	**6244.38** 	_41.23 	226.80 (2) 	653_ 	insecure, 100% bias, collisions, distrib, BIC, machine-specific (SSE4.2/NEON)
> [crc32_hw1](https://github.com/rurban/smhasher/blob/master/doc/crc32_hw1.txt) 	**7569.29** 	_49.07 	233.75 (3) 	671_ 	insecure, 100% bias, collisions, distrib, BIC, machine-specific (x86 SSE4.2)
> [crc64_hw](https://github.com/rurban/smhasher/blob/master/doc/crc64_hw.txt) 	**6143.62** 	_40.48 	223.13 (2) 	652_ 	insecure, 100% bias, collisions, distrib, BIC, machine-specific (SSE4.2/NEON)

kirillt:
- #65
- #73

PR #73, opened 2024-01-24
#65: Replace `CRC32` with `BLAKE3`  Hash Function

tareknaser:
### Description
This pull request updates the hashing method for the `ResourceId` struct, replacing `CRC32` with `BLAKE3` hash function.
Related Issue: #65
### Changes 
The changes include:
- Addition of benchmarks for the `ResourceId::compute_bytes` method using the `criterion` crate with the `html_reports` feature enabled. Run `cargo bench` to generate benchmarks in the `/target/criterion` folder
-  Removing collision handling as `BLAKE3` is a cryptographic hash function

> Refer to individual commit messages and bodies for detailed information

### DIFF
--- a/src/id.rs
+++ b/src/id.rs
@@ -1,4 +1,6 @@
 use anyhow::anyhow;
+use base64::engine::general_purpose::STANDARD as BASE64;
+use base64::engine::Engine as _;
 use blake3::Hasher as Blake3Hasher;
 use log;
 use serde::{Deserialize, Serialize};
@@ -24,13 +26,13 @@ use crate::{ArklibError, Result};
     Serialize,
 )]
 pub struct ResourceId {
-    pub data_size: u64,
     pub blake3: [u8; 32],
 }
 
 impl Display for ResourceId {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        write!(f, "{}-{:?}", self.data_size, self.blake3)
+        let blake3_str = BASE64.encode(&self.blake3);
+        write!(f, "{}", blake3_str)
     }
 }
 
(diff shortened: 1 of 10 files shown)


kirillt:
Let's finish the benchmarks first:
- #74 

After we have index benchmark, we'll be able to judge better about performance penalty of Blake3. Collision tracking, necessary for CRC32, could eat some performance up, too.

src/id.rs
```diff
@@ -25,12 +25,12 @@ use crate::{ArklibError, Result};
...
 pub struct ResourceId {
     pub data_size: u64,
-    pub crc32: u32,
+    pub blake3: [u8; 32],
 }
 
 impl Display for ResourceId {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        write!(f, "{}-{}", self.data_size, self.crc32)
+        write!(f, "{}-{:?}", self.data_size, self.blake3)
```
kirillt, line 33: Something like https://crates.io/crates/base64 should be used instead of `{:?}`.

src/id.rs
```diff
@@ -40,9 +40,19 @@ impl FromStr for ResourceId {
...
+        if bytes.len() != 32 {
+            return Err(ArklibError::Parse);
+        }
+        let mut blake3 = [0u8; 32];
+        for (i, byte_str) in bytes.iter().enumerate() {
+            blake3[i] = byte_str
+                .trim()
+                .parse()
+                .map_err(|_e| ArklibError::Parse)?;
+        }
```
kirillt, line 54: Something like https://crates.io/crates/base64 should be used instead of `{:?}`.

src/id.rs
```diff
@@ -25,12 +25,12 @@ use crate::{ArklibError, Result};
 )]
 pub struct ResourceId {
     pub data_size: u64,
```
kirillt, line 27: Actually, we should remove `data_size`, too.

tareknaser:
This PR should be rewritten when https://github.com/ARK-Builders/arklib/pull/90 is merged here and into `ark-rust` as well. 

I can re rewrite the logic in `ark-rust` directly so that it implements `ResourceIdTrait`

tareknaser:
Closing this in favor of https://github.com/ARK-Builders/ark-rust/pull/32

PR #74, opened 2024-01-26
Add Hash Function Benchmarks

tareknaser:
### Description
This PR adds benchmarking infrastructure using `criterion` crate. For now, a single benchmark is added.
Related issue: #65 
### Changes
The file `benches/hash_benchmark.rs` includes the `compute_bytes_benchmark` function, which has three benchmarks for different use cases:
- `compute_bytes_small`: Benchmarks with small input data.
- `compute_bytes_medium`: Evaluates medium-sized input data.
- `compute_bytes_large`: Measures large input data performance.

`criterion` is configured as a dev dependency with the `html_reports` feature for detailed HTML reports in `target/criterion/`.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,0 +1,19 @@
+name: Benchmark Pull Requests
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  run_benchmarks:
+    name: Run Benchmarks
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - uses: boa-dev/criterion-compare-action@v3
+        with:
+          branchName: ${{ github.base_ref }}
+          token: ${{ secrets.GITHUB_TOKEN }}
--- a/benches/compute_bytes_benchmark.rs
+++ b/benches/compute_bytes_benchmark.rs
@@ -8,7 +8,7 @@ fn generate_random_data(size: usize) -> Vec<u8> {
     (0..size).map(|_| rng.gen()).collect()
 }
 
-fn compute_bytes_benchmark(c: &mut Criterion) {
+fn compute_bytes_on_raw_data(c: &mut Criterion) {
     let inputs = [
         ("compute_bytes_small", 1024),
         ("compute_bytes_medium", 8192),
@@ -52,7 +52,7 @@ fn compute_bytes_on_files_benchmark(c: &mut Criterion) {
 
 criterion_group!(
     benches,
-    compute_bytes_benchmark,
+    compute_bytes_on_raw_data,
     compute_bytes_on_files_benchmark
 );
 criterion_main!(benches);
--- a/benches/compute_bytes_benchmark.rs
+++ b/benches/compute_bytes_benchmark.rs
@@ -11,42 +11,51 @@ fn generate_random_data(size: usize) -> Vec<u8> {
 fn compute_bytes_on_raw_data(c: &mut Criterion) {
     let inputs = [
         ("compute_bytes_small", 1024),
-        ("compute_bytes_medium", 8192),
-        ("compute_bytes_large", 65536),
+        ("compute_bytes_medium", 65536),
+        ("compute_bytes_large", 1048576),
     ];
 
     for (name, size) in inputs.iter() {
         let input_data = generate_random_data(*size);
-        c.bench_function(name, move |b| {
+        let mut group = c.benchmark_group(name.to_string());
+        group.measurement_time(std::time::Duration::from_secs(5)); // Set the measurement time here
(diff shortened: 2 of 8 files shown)


kirillt:
Looks great. Could you also implement these points?
- [ ] run benchmark during our GitHub Actions pipeline
- [ ] increase data sizes for hash benchmark
- [ ] add "fixed path on filesystem" case for hash benchmark
    e.g. `tests/lena.jpg` can be used, anyone will be able to replace it locally to run benchmark on a huge file
- [ ] add index construction benchmark
- [ ] add "fixed path on filesystem" case for index benchmark
    e.g. `tests` folder, which could be filled with diverse data locally

tareknaser:
I've set up benchmarks to run automatically on every pull request to the main branch using `boa-dev/criterion-compare-action`.
This configuration runs benchmarks on both the pull request branch and the main branch, and then posts the results as a comment on the pull request. You can see an example of how it looks when merged in the following link: [Example](https://github.com/boa-dev/criterion-compare-action?tab=readme-ov-file#example)

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,19 @@
+name: Benchmark Pull Requests
```
kirillt, line 1: Let's name the workflow simpler, e.g. something like `Run benchmarks`.

Also let's name the file as `benchmark.yml`, because later, the other workflow might be set for PRs, too.

kirillt:
Can we catch benchmarks errors, too? It's not critical but would be nice.

E.g. I modify the path to something on my filesystem:
```
[kirill@lenovo arklib]$ git diff benches/index_build_benchmark.rs
diff --git a/benches/index_build_benchmark.rs b/benches/index_build_benchmark.rs
index f3fd97b..8df6959 100644
--- a/benches/index_build_benchmark.rs
+++ b/benches/index_build_benchmark.rs
@@ -2,7 +2,7 @@ use arklib::index::ResourceIndex;
 use criterion::{black_box, criterion_group, criterion_main, Criterion};
 
 fn index_build_benchmark(c: &mut Criterion) {
-    let path = "tests/"; // Set the path to the directory containing the resources
+    let path = "/mnt/data/tests"; // Set the path to the directory containing the resources
 
     c.bench_function("index_build", move |b| {
         b.iter(|| {
```
But it doesn't exist:
```
[kirill@lenovo arklib]$ ls /mnt/data/tests
ls: cannot access '/mnt/data/tests': No such file or directory
```
Right now, the benchmark just passes quickly:
```
[kirill@lenovo arklib]$ RUST_LOG=debug cargo bench index_build
    Finished bench [optimized] target(s) in 0.25s
     Running benches/compute_bytes_benchmark.rs (target/release/deps/compute_bytes_benchmark-1ca61f88413531a5)
Gnuplot not found, using plotters backend
     Running benches/index_build_benchmark.rs (target/release/deps/index_build_benchmark-3021d4b85891b6cc)
Gnuplot not found, using plotters backend
index_build             time:   [33.091 µs 33.677 µs 34.275 µs]
                        change: [-6.2644% -3.3005% -0.2257%] (p = 0.04 < 0.05)
                        Change within noise threshold.
Found 7 outliers among 100 measurements (7.00%)
  3 (3.00%) low mild
  4 (4.00%) high mild
```

It's kinda expected, but if we can catch this error, it would be great.

benches/compute_bytes_benchmark.rs
```diff
@@ -0,0 +1,58 @@
...
+fn generate_random_data(size: usize) -> Vec<u8> {
+    let mut rng = rand::thread_rng();
+    (0..size).map(|_| rng.gen()).collect()
+}
+
+fn compute_bytes_on_raw_data(c: &mut Criterion) {
+    let inputs = [
+        ("compute_bytes_small", 1024),
+        ("compute_bytes_medium", 8192),
+        ("compute_bytes_large", 65536),
```
kirillt, line 15: Let's set `medium` to `65536` (1mb) and `large` to `1048576` (16mb)

benches/index_build_benchmark.rs
```diff
@@ -1,14 +1,37 @@
...
+            b.iter(|| {
+                ResourceIndex::build(black_box(path.to_string()));
+            });
+        },
+    );
+
+    group.finish();
+
+    // Print the number of collisions (i.e. resources with the same hash)
+    let index = ResourceIndex::build(path.to_string());
```
kirillt, line 32: Let's reuse the same index instance from the benchmark runs.

This way, the control of runs number would be more precise.. Imagine huge folder on local filesystem.

kirillt (approved):

benches/index_build_benchmark.rs
tareknaser: good point!

Issue #86, opened 2024-02-16
Store relative paths in the index mappings


kirillt:
* Only relative paths should be used in the internal index mappings
* Canonicalization must be performed before registering the paths
* The common prefix ("root") should be trimmed from the paths
* The lib consumers must be able to reconstruct absolute paths

tareknaser:
We can close this issue in favor of https://github.com/ARK-Builders/ark-rust/issues/20

PR #87, opened 2024-02-16
Refactor `src/index.rs` and Enhance Documentation

### LINKED ISSUE #89: Empty index is generated if name of the root folder starts with `.`
Hidden subfolders must be ignored during the filesystem traversal. But if the root is hidden we still need to perform the traversal.

tareknaser:
## Description
This PR focuses on refactoring the `src/index.rs` file. 
## Changes
The changes primarily include:
- Define constants at the beginning of the file
- Add doc comments for structs, fields, and methods 
- Use `serde::{Deserialize, Serialize}` for loading and storing of the index file
- Rename the `discover_paths` function to `discover_files` 
- Refactor certain parts of the code to improve readability, including reducing indentation where possible
- Add a new test to ensure that only canonical paths are present in the index
- Use of `TempDir` for testing 
- Updating platform conditions for improved cross-compatibility (`#[cfg(target_family = "unix")]`)

## Edit
Fixes #89 with 633335ab42a603fe646b1bfe7b687a271b7ab9b3

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -593,6 +593,9 @@ fn discover_files<P: AsRef<Path>>(root_path: P) -> HashMap<PathBuf, DirEntry> {
     discovered_files
 }
 
+/// Scans a single file entry and extracts its metadata to create an index entry
+///
+/// Returns an error if the path is a directory or if the file is empty
 fn scan_entry(path: &Path, metadata: Metadata) -> Result<IndexEntry> {
     if metadata.is_dir() {
         return Err(ArklibError::Path("Path is expected to be a file".into()));
@@ -600,10 +603,7 @@ fn scan_entry(path: &Path, metadata: Metadata) -> Result<IndexEntry> {
 
     let size = metadata.len();
     if size == 0 {
-        Err(std::io::Error::new(
-            std::io::ErrorKind::InvalidData,
-            "Empty resource",
-        ))?;
+        return Err(ArklibError::Path("Empty file".into()));
     }
 
     let id = ResourceId::compute(size, path)?;
@@ -612,6 +612,9 @@ fn scan_entry(path: &Path, metadata: Metadata) -> Result<IndexEntry> {
     Ok(IndexEntry { id, modified })
 }
 
+/// Scans multiple file entries and creates index entries for each one
+///
+/// Returns a hashmap of file paths to their corresponding index entries
 fn scan_entries(
     entries: HashMap<PathBuf, DirEntry>,
 ) -> HashMap<PathBuf, IndexEntry> {
--- a/src/index.rs
+++ b/src/index.rs
@@ -609,7 +609,7 @@ impl ResourceIndex {
                 "Any collision must involve at least 2 resources"
             );
             *collisions -= 1;
-            if *collisions <= 1 {
+            if *collisions == 1 {
                 self.collisions.remove(&old_id);
             }
 
--- a/src/index.rs
+++ b/src/index.rs
@@ -1,23 +1,25 @@
 use anyhow::anyhow;
-use canonical_path::{CanonicalPath, CanonicalPathBuf};
 use itertools::Itertools;
 use log;
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
 use std::collections::{HashMap, HashSet};
 use std::fs::{self, File, Metadata};
-use std::io::{BufRead, BufReader, Write};
-use std::ops::Add;
+use std::io::BufReader;
+use std::io::BufWriter;
 use std::path::{Path, PathBuf};
-use std::str::FromStr;
-use std::time::{Duration, SystemTime, UNIX_EPOCH};
+use std::time::{Duration, SystemTime};
 use walkdir::{DirEntry, WalkDir};
 
 use crate::{id::ResourceId, ArklibError, Result, ARK_FOLDER, INDEX_PATH};
 
 pub const RESOURCE_UPDATED_THRESHOLD: Duration = Duration::from_millis(1);
-pub type Paths = HashSet<CanonicalPathBuf>;
+pub type Paths = HashSet<PathBuf>;
 
 /// IndexEntry represents a [`ResourceId`] and the time it was last modified
-#[derive(Eq, Ord, PartialEq, PartialOrd, Hash, Clone, Debug)]
+#[derive(
+    Eq, Ord, PartialEq, PartialOrd, Hash, Clone, Debug, Serialize, Deserialize,
+)]
 pub struct IndexEntry {
     /// The time the resource was last modified
     pub modified: SystemTime,
@@ -30,12 +32,14 @@ pub struct IndexEntry {
 /// This struct maintains mappings between resource IDs and their corresponding
 /// file paths, as well as mappings between file paths and index entries
 /// Additionally, it keeps track of collisions that occur during indexing
-#[derive(PartialEq, Clone, Debug)]
+#[serde_as]
+#[derive(PartialEq, Clone, Debug, Serialize, Deserialize)]
 pub struct ResourceIndex {
     /// A mapping of resource IDs to their corresponding file paths
-    pub id2path: HashMap<ResourceId, CanonicalPathBuf>,
+    #[serde_as(as = "Vec<(_, _)>")]
+    pub id2path: HashMap<ResourceId, PathBuf>,
     /// A mapping of file paths to their corresponding index entries
-    pub path2id: HashMap<CanonicalPathBuf, IndexEntry>,
+    pub path2id: HashMap<PathBuf, IndexEntry>,
     /// A mapping of resource IDs to the number of collisions they have
     pub collisions: HashMap<ResourceId, usize>,
     /// The root path of the index
@@ -51,7 +55,7 @@ pub struct IndexUpdate {
     /// Set of resource IDs that have been deleted
     pub deleted: HashSet<ResourceId>,
     /// Map of file paths to resource IDs that have been added
-    pub added: HashMap<CanonicalPathBuf, ResourceId>,
+    pub added: HashMap<PathBuf, ResourceId>,
 }
 
 impl ResourceIndex {
@@ -90,95 +94,42 @@ impl ResourceIndex {
         index
     }
 
+    /// Loads a previously stored resource index from the root path
+    ///
+    /// This function reads the index from the file system and returns a new
+    /// [`ResourceIndex`] instance. It looks for the index fie in
+    /// `root_path/.ark/index`
     pub fn load<P: AsRef<Path>>(root_path: P) -> Result<Self> {
-        let root_path: PathBuf = root_path.as_ref().to_owned();
+        let index_path = root_path
+            .as_ref()
+            .join(ARK_FOLDER)
+            .join(INDEX_PATH);
+        log::info!("Loading the index from file: {}", index_path.display());
 
-        let index_path: PathBuf = root_path.join(ARK_FOLDER).join(INDEX_PATH);
-        log::info!("Loading the index from file {}", index_path.display());
-        let file = File::open(&index_path)?;
-        let mut index = ResourceIndex {
-            id2path: HashMap::new(),
-            path2id: HashMap::new(),
-            collisions: HashMap::new(),
-            root: root_path.clone(),
-        };
-
-        // We should not return early in case of missing files
-        let lines = BufReader::new(file).lines();
-        for line in lines {
-            let line = line?;
-
-            let mut parts = line.split(' ');
-
-            let modified = {
-                let str = parts.next().ok_or(ArklibError::Parse)?;
-                UNIX_EPOCH.add(Duration::from_millis(
-                    str.parse().map_err(|_| ArklibError::Parse)?,
-                ))
-            };
-
-            let id = {
-                let str = parts.next().ok_or(ArklibError::Parse)?;
-                ResourceId::from_str(str)?
-            };
-
-            let path: String =
-                itertools::Itertools::intersperse(parts, " ").collect();
-            let path: PathBuf = root_path.join(Path::new(&path));
-            match CanonicalPathBuf::canonicalize(&path) {
-                Ok(path) => {
-                    log::trace!("[load] {} -> {}", id, path.display());
-                    index.insert_entry(path, IndexEntry { id, modified });
-                }
-                Err(_) => {
-                    log::warn!("File {} not found", path.display());
-                    continue;
-                }
-            }
-        }
+        let file = File::open(index_path)?;
+        let reader = BufReader::new(file);
+        let index: ResourceIndex = serde_json::from_reader(reader)?;
 
         Ok(index)
     }
 
+    /// Stores the resource index to the file system
+    ///
+    /// This function writes the index to the file system. It writes the index
+    /// to `root_path/.ark/index` and creates the directory if it does not exist
     pub fn store(&self) -> Result<()> {
         log::info!("Storing the index to file");
-
         let start = SystemTime::now();
 
-        let index_path = self
-            .root
-            .to_owned()
-            .join(ARK_FOLDER)
-            .join(INDEX_PATH);
-
-        let ark_dir = index_path.parent().unwrap();
-        fs::create_dir_all(ark_dir)?;
-
-        let mut file = File::create(index_path)?;
-
-        let mut path2id: Vec<(&CanonicalPathBuf, &IndexEntry)> =
-            self.path2id.iter().collect();
-        path2id.sort_by_key(|(_, entry)| *entry);
-
-        for (path, entry) in path2id.iter() {
-            log::trace!("[store] {} by path {}", entry.id, path.display());
-
-            let timestamp = entry
-                .modified
-                .duration_since(UNIX_EPOCH)
-                .map_err(|_| {
-                    ArklibError::Other(anyhow!("Error using duration since"))
-                })?
-                .as_millis();
-
-            let path =
-                pathdiff::diff_paths(path.to_str().unwrap(), self.root.clone())
-                    .ok_or(ArklibError::Path(
-                        "Couldn't calculate path diff".into(),
-                    ))?;
-
-            writeln!(file, "{} {} {}", timestamp, entry.id, path.display())?;
+        let ark_folder = self.root.join(ARK_FOLDER);
+        if !ark_folder.exists() {
+            fs::create_dir(ark_folder.clone())?;
         }
+        let index_path = ark_folder.join(INDEX_PATH);
+        let file = File::create(index_path)?;
+
+        let writer = BufWriter::new(file);
+        serde_json::to_writer(writer, self)?;
 
         log::trace!(
             "Storing the index took {:?}",
@@ -237,10 +188,10 @@ impl ResourceIndex {
             .cloned()
             .collect();
 
-        let created_paths: HashMap<CanonicalPathBuf, DirEntry> = curr_entries
+        let created_paths: HashMap<PathBuf, DirEntry> = curr_entries
             .iter()
             .filter_map(|(path, entry)| {
-                if !preserved_paths.contains(path.as_canonical_path()) {
+                if !preserved_paths.contains(path) {
                     Some((path.clone(), entry.clone()))
                 } else {
                     None
@@ -249,10 +200,10 @@ impl ResourceIndex {
             .collect();
 
         log::debug!("Checking updated paths");
-        let updated_paths: HashMap<CanonicalPathBuf, DirEntry> = curr_entries
+        let updated_paths: HashMap<PathBuf, DirEntry> = curr_entries
             .into_iter()
             .filter(|(path, dir_entry)| {
-                if !preserved_paths.contains(path.as_canonical_path()) {
+                if !preserved_paths.contains(path) {
                     false
                 } else {
                     let our_entry = &self.path2id[path];
@@ -314,9 +265,7 @@ impl ResourceIndex {
             .cloned()
             .chain(updated_paths.keys().cloned())
             .for_each(|path| {
-                if let Some(entry) =
-                    self.path2id.remove(path.as_canonical_path())
-                {
+                if let Some(entry) = self.path2id.remove(&path) {
                     let k = self.collisions.remove(&entry.id).unwrap_or(1);
                     if k > 1 {
                         self.collisions.insert(entry.id, k - 1);
@@ -334,15 +283,14 @@ impl ResourceIndex {
                 }
             });
 
-        let added: HashMap<CanonicalPathBuf, IndexEntry> =
-            scan_entries(updated_paths)
-                .into_iter()
-                .chain({
-                    log::debug!("Checking added paths");
-                    scan_entries(created_paths).into_iter()
-                })
-                .filter(|(_, entry)| !self.id2path.contains_key(&entry.id))
-                .collect();
+        let added: HashMap<PathBuf, IndexEntry> = scan_entries(updated_paths)
+            .into_iter()
+            .chain({
+                log::debug!("Checking added paths");
+                scan_entries(created_paths).into_iter()
+            })
+            .filter(|(_, entry)| !self.id2path.contains_key(&entry.id))
+            .collect();
 
         for (path, entry) in added.iter() {
             if deleted.contains(&entry.id) {
@@ -358,7 +306,7 @@ impl ResourceIndex {
             self.insert_entry(path.clone(), entry.clone());
         }
 
-        let added: HashMap<CanonicalPathBuf, ResourceId> = added
+        let added: HashMap<PathBuf, ResourceId> = added
             .into_iter()
             .map(|(path, entry)| (path, entry.id))
             .collect();
@@ -378,8 +326,8 @@ impl ResourceIndex {
             ));
         }
 
-        let path_buf = CanonicalPathBuf::canonicalize(path)?;
-        let path = path_buf.as_canonical_path();
+        let path_buf = fs::canonicalize(path)?;
+        let path = path_buf.as_path();
 
         return match fs::metadata(path) {
             Err(_) => {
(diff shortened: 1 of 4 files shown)


src/index.rs
```diff
@@ -1,65 +1,90 @@
...
 
+    /// Builds a new resource index from scratch using the root path
+    ///
+    /// This function recursively scans the directory structure starting from
+    /// the root path, constructs index entries for each resource found, and
+    /// populates the resource index
     pub fn build<P: AsRef<Path>>(root_path: P) -> Self {
-        log::info!("Building the index from scratch");
-        let root_path: PathBuf = root_path.as_ref().to_owned();
+        let root_path = root_path.as_ref().to_owned();
```
kirillt, line 74: We also should canonicalize it.

src/index.rs
```diff
@@ -1,65 +1,90 @@
...
-pub type Paths = HashSet<CanonicalPathBuf>;
-
 impl ResourceIndex {
+    /// Returns the number of entries in the index
+    ///
+    /// Note that the actual size is lower in presence of collisions
     pub fn size(&self) -> usize {
-        //the actual size is lower in presence of collisions
         self.path2id.len()
     }
```
kirillt, line 66: Do you think separate functions is a good idea?
1. `pub fn count_files() -> usize` which equals to `self.path2id.len()`
2. `pub fn count_resources() -> usize` which equals to `self.id2path.len()`

src/index.rs
```diff
@@ -68,95 +93,42 @@ impl ResourceIndex {
...
-                }
-                Err(_) => {
-                    log::warn!("File {} not found", path.display());
-                    continue;
-                }
-            }
-        }
+        let file = File::open(index_path)?;
+        let reader = BufReader::new(file);
+        let index: ResourceIndex = serde_json::from_reader(reader)?;
```
kirillt, line 110: Am I right that this changes the current file format to JSON?

In the `main` branch, index is stored in plain-text format:
```
1675087824710 58922-3276384608 Info/crdt_vs_ot.jpg
1675087824610 125308-3041567246 Info/Watson-Crick & Hoogsteen.png
1675087831180 386929-63288935 Screenshots/Screenshot_20210717_120856.jpg
1675087831440 945565-3484272718 Screenshots/Screenshot_20220111_040205_com.adobe.reader.jpg
1675087832200 57186-2804307164 Screenshots/lenovo-fedora-20210427-023231-183683953.png
1675087832990 107862-4031618865 Screenshots/lenovo-fedora-20210328-050520-182040364.png
1675087833830 343648-561346264 Screenshots/Screenshot_20210804_003308.jpg
1675087834130 257481-490338418 Info/Technologies.jpg
1675087846130 415620-3764250547 Gif/сitrus.gif
```

src/index.rs
```diff
@@ -68,95 +93,42 @@ impl ResourceIndex {
...
-            writeln!(file, "{} {} {}", timestamp, entry.id, path.display())?;
+        let ark_folder = self.root.join(ARK_FOLDER);
+        if !ark_folder.exists() {
+            fs::create_dir(ark_folder.clone())?;
         }
+        let index_path = ark_folder.join(INDEX_PATH);
+        let file = File::create(index_path)?;
+
+        let writer = BufWriter::new(file);
+        serde_json::to_writer(writer, self)?;
```
kirillt, line 131: The file format from `main` branch is also sorted by timestamps. This is useful for keeping the index file same on different devices and between runs, and also convenient for debugging.

src/index.rs
```diff
@@ -167,46 +139,46 @@ impl ResourceIndex {
...
                 Ok(Self::build(root_path))
             }
         }
     }
 
+    /// Updates the index based on the current state of the file system
+    ///
+    /// Returns an [`IndexUpdate`] object containing the paths of deleted and
+    /// added resources
     pub fn update_all(&mut self) -> Result<IndexUpdate> {
```
kirillt, line 175: @kirillt compare with https://github.com/ARK-Builders/arklib/blob/index/update-one/src/index.rs#L210

src/index.rs
```diff
@@ -332,72 +303,82 @@ impl ResourceIndex {
...
+    ///
+    /// The caller must ensure that:
+    /// - The index is up-to-date except for this single path
+    /// - The path hasn't been indexed before
+    ///
+    /// Returns an error if:
+    /// - The path does not exist
+    /// - Metadata retrieval fails
     pub fn index_new(&mut self, path: &dyn AsRef<Path>) -> Result<IndexUpdate> {
-        log::debug!("Indexing a new path");
```
kirillt, line 351: Let's use name `track_addition` if it's same function as here https://github.com/ARK-Builders/arklib/blob/index/update-one/src/index.rs#L359

src/index.rs
```diff
@@ -332,72 +303,82 @@ impl ResourceIndex {
...
+    /// * the path has been indexed before
+    /// * the path maps into `old_id`
+    /// * the content by the path has been modified
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the path does not exist, if the path is a directory
+    /// or an empty file, if the index cannot find the specified path, or if
+    /// the content of the path has not been modified.
     pub fn update_one(
```
kirillt, line 382: It seems to me, that we should separate this method into `track_deletion` and `track_update` as it was done in #72

https://github.com/ARK-Builders/arklib/blob/index/update-one/src/index.rs#L393
https://github.com/ARK-Builders/arklib/blob/index/update-one/src/index.rs#L416

The idea is that `track` methods serve as performance workaround when end-users know what they are doing. On other hand, `update_all` should be efficient enough to be called all the time. But from API point of view, it `track` methods are also simpler sometimes.

OPEN FOR DISCUSSION

src/index.rs
```diff
@@ -516,9 +471,41 @@ impl ResourceIndex {
         self.path2id.insert(path, entry);
     }
 
+    /// Removes the given resource ID from the index and returns an update
+    /// containing the deleted entries
+    pub fn forget_id(&mut self, old_id: ResourceId) -> Result<IndexUpdate> {
```
kirillt, line 476: Same as this? https://github.com/ARK-Builders/arklib/blob/index/update-one/src/index.rs#L393

src/index.rs
```diff
@@ -574,55 +554,66 @@ impl ResourceIndex {
...
             Ok(entry) => {
-                let path = entry.path();
+                let path = entry.path().to_path_buf();
                 if !entry.file_type().is_dir() {
-                    match CanonicalPathBuf::canonicalize(path) {
-                        Ok(canonical_path) => Some((canonical_path, entry)),
+                    // canonicalize the path to avoid duplicates
+                    match fs::canonicalize(&path) {
+                        Ok(canonical_path) => {
+                            discovered_files.insert(canonical_path, entry);
```
kirillt, line 585: After canonicalization, we should trim the common prefix ("root").

src/index.rs
```diff
@@ -656,26 +650,18 @@ fn scan_entries(
...
-        .file_name()
-        .to_str()
-        .map(|s| s.starts_with('.'))
-        .unwrap_or(false)
-}
-
 #[cfg(test)]
 mod tests {
+    use super::fs;
     use crate::id::ResourceId;
```
kirillt, line 656: @kirillt look into the test cases later.

src/index.rs
```diff
@@ -1,65 +1,105 @@
...
+///
+/// This struct maintains a mapping from resource IDs to their corresponding
+/// file paths, as well as the mapping of the opposite direction from file
+/// paths to index entries, which are simply resource IDs with modification
+/// timestamps.
+///
+/// The paths stored in the mappings are not absolute paths, but are relative
+/// to the "root". See methods `TODO` and `TODO` in the API, which can provide
+/// the end users with absolute/canonical(TODO) paths for the convenince.
+/// Otherwise, the end user is responsible for managing the working dir
```
kirillt, line 40: @tareknaser do you think we need functions to get absolute paths?

Actually, we could forbid querying particular paths because, ideally, the lib must be used in FRP way, i.e. only react to updates, which could be presented as a stream in the library wrappers. But it could be a bit tough for the end-users... If we adopt this way, the API would become cleaner though. We would need initialization parameter which specifies what kind of paths we emit in the updates: relative or absolute. This would become the preference of the end-user.

If we provide only "reactive API", then I'm not sure what to do with `track` functions.

kirillt:
@tareknaser could you also check this issue?
- #89

src/index.rs
tareknaser: Right. Thanks for the catch

src/index.rs
```diff
@@ -1,65 +1,90 @@
...
-
+    /// A mapping of resource IDs to their corresponding file paths
+    #[serde_as(as = "Vec<(_, _)>")]
+    pub id2path: HashMap<ResourceId, PathBuf>,
+    /// A mapping of file paths to their corresponding index entries
+    pub path2id: HashMap<PathBuf, IndexEntry>,
+    /// A mapping of resource IDs to the number of collisions they have
     pub collisions: HashMap<ResourceId, usize>,
+    /// The root path of the index
     root: PathBuf,
```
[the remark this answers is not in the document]
tareknaser: I think the recommended API for building the index should be using `ResourceIndex::build()`, which handles path canonicalization internally. 
If we make all `ResourceIndex` fields public, we allow the creation of objects with inconsistent or invalid data, which might result in unpredictable behavior.

src/index.rs
tareknaser: I have found that storing and loading the same index with the `main` branch doesn't produce the same object. Not a matter of sorting but timestamps might be different as well. The code in the `main` branch for storing and loading contains more logic that just writing the struct to a file.
The idea here was to make `ResourceIndex` implement `Serialize`. with this setup, [we can serialize to various formats](https://serde.rs/#data-formats)

but It's up to you, we can revert to the old file format if you prefer it

src/index.rs
tareknaser: Yes. good point
we can sort before serializing.

src/index.rs
tareknaser: I didn't really change the function logic. Should I rename it to `track_deletion`?

src/index.rs
tareknaser: Ok so the API will look like:
- `track_addition()`, `track_update()` and `track_deletion()` when the user wants to be specific about the action
- `update_all()` otherwise

This looks like a clean API but I have some questions about other public methods
- what are the use cases for `ResourceIndex::provide()`?
- we can now remove `forget_id()` in favor of `track_deletion()` or they do different tasks?

src/index.rs
tareknaser: That's related to #86. Do you want me to implement it in this PR?

src/index.rs
tareknaser: I think the default should always be that we are dealing with relative path. The `root_path` is a field in `ResourceIndex` and any relative path can be prepended with it to get the canonical path. 

We just have to make sure that `root_path` is canonical on its own. We currently do this in `ResourceIndex::build()` but we also need a getter for it

src/index.rs
kirillt: Probably we should make the mappings private.

src/index.rs
kirillt: I like your idea, I just don't want to write huge JSON to disk.

Some compact format would be great: postcard, CBOR, what else could serve this purpose?

kirillt: On other hand, if we use automatic serialization, then the format will be excessive most likely: 2 mappings means creating two separate structures. Right now, we write only one table and derive 2 mappings from it during reading.

src/index.rs
kirillt: Ah yeah, we can do it in separate PR.

src/index.rs
```diff
@@ -586,6 +586,7 @@ fn discover_files<P: AsRef<Path>>(root_path: P) -> HashMap<PathBuf, DirEntry> {
 
     let mut discovered_files = HashMap::new();
     let walker = WalkDir::new(root_path)
+        .min_depth(1)
```
kirillt, line 589: :+1:

src/index.rs
tareknaser: Updated the maps of `ResourceIndex` to be private. 
I also added 2 methods, `ResourceIndex::count_files()` and `ResourceIndex::count_resources()` to return the size of `path2id` and `id2path` respectively

src/index.rs
tareknaser: > On other hand, if we use automatic serialization, then the format will be excessive 

Good point. I noticed that as well.

I now reverted `ResourceIndex::store()` and `ResourceIndex::load()` to the old code. I just kept the code to make them serializable to have this possibility later. 

I just wanted to add some notes about the current implementation of the 2 methods:
- we currently don't store or load information about collisions in `ResourceIndex`
- I noticed some inconsistencies when trying to store an object of `ResourceIndex` and then load it. I noticed some differences not only in order but also in `modified_time`. I think this may be a problem with how we parse the time

I could move this conversation to an issue and work on a separate PR to solve it

src/index.rs
kirillt: Agree, and also we must ensure that the paths are canonicalized before trimming the common prefix.

src/index.rs
```diff
@@ -1,65 +1,116 @@
...
+/// This struct maintains a mapping from resource IDs to their corresponding
+/// file paths, as well as the mapping of the opposite direction from file
+/// paths to index entries, which are simply resource IDs with modification
+/// timestamps.
+///
+/// The paths stored in the mappings are not absolute paths, but are relative
+/// to the "root". See methods `TODO` and `TODO` in the API, which can provide
+/// the end users with absolute/canonical(TODO) paths for the convenince.
+/// Otherwise, the end user is responsible for managing the working dir
+/// and relative paths resolution.
```
kirillt, line 45: In the end, our API should consist of 2 halves:
* "Snapshot API" allowing to query particular paths or ids, separate functions for relative and absolute paths. We could also implement something like Iter interface.
* "Reactive API" allowing updates handling without explicit queries. It can be done in pull-based manner like now (update_all), push-based using Tokio streams and/or registering handlers which would be called automatically (variant of push-based) 

We should discuss and create new issues for this.

src/index.rs
kirillt: @tareknaser if I remember correctly:
* Function `forget_id` is only utilitary, became private `forget_entry` in the other branch of the codebase. It's dual to `insert_entry`, both are called from `track` methods.
* `ResourceIndex::provide()` is the starting point for any user/app. It combines index loading and updating in single step. This could be done by separate calls, but it seems to me that users will always need updated index before they do something. I can imagine rare use-cases where they need more control, so we should expose separate steps, too.

src/index.rs
kirillt: @tareknaser this is interesting
> I noticed some inconsistencies when trying to store an object of ResourceIndex and then load it. I noticed some differences not only in order but also in modified_time. I think this may be a problem with how we parse the time

Could you tell more about the timestamps discrepancies? Probably steps to reproduce with `ark-cli`? Yeah, we could create separate issue if this is a bug indeed.

> we currently don't store or load information about collisions in ResourceIndex

Yes, but is it an issue? Collisions should be derivable from the table itself, right?

src/index.rs
kirillt: All good here. The code actually became more readable :+1:

src/index.rs
kirillt: Btw test case ` fn index_entry_order()` could help debug lines ordering.

src/index.rs
```diff
@@ -516,9 +563,41 @@ impl ResourceIndex {
...
+    /// Removes an entry with the specified path and updates the collision
+    /// information accordingly
+    ///
+    /// Returns an update containing the deleted entries
     fn forget_path(
         &mut self,
-        path: &CanonicalPath,
+        path: &Path,
         old_id: ResourceId,
     ) -> Result<IndexUpdate> {
```
kirillt, line 602: 1. I would return just `deleted` collection, because `added` is always empty here.
2. Maybe name it `forget_entry` since path and id have same importance?

src/index.rs
```diff
@@ -529,36 +608,29 @@ impl ResourceIndex {
                 "Any collision must involve at least 2 resources"
             );
             *collisions -= 1;
-
-            if *collisions == 1 {
+            if *collisions <= 1 {
```
kirillt, line 611: Shouldn't `collisions == 0` be treated as a broken invariant?

src/index.rs
```diff
@@ -516,9 +563,41 @@ impl ResourceIndex {
...
+        self.id2path.remove(&old_id);
+
+        let mut deleted = HashSet::new();
+        deleted.insert(old_id);
+
+        Ok(IndexUpdate {
+            added: HashMap::new(),
+            deleted,
+        })
+    }
```
kirillt, line 592: From `index/update-one` branch:
```
    /// The caller must ensure that:
    /// * the index is up-to-date except this single id
    /// * the resource with this id has been indexed before
    /// * the resource with this id doesn't exist anymore
    /// 
    /// Should only be used if reactive updates are not possible.
    pub fn track_deletion(&mut self, id: ResourceId) -> Result<IndexUpdate> {
        log::debug!("Tracking a single deletion in the index");

        let indexed_path = self.id2path.get(&id);
        if indexed_path.is_none() {
            return Err(ArklibError::Path(
                "The id cannot be found in the index".into(),
            )); 
        }   

        let indexed_path = indexed_path.unwrap().clone();
        self.forget_entry(indexed_path.as_canonical_path(), id)?;

        Ok(IndexUpdate::default().deleted(id))
    }   
```

src/index.rs
tareknaser: > Could you tell more about the timestamps discrepancies? Probably steps to reproduce with ark-cli? Yeah, we could create separate issue if this is a bug indeed.

I actually noticed this difference when I wrote a simple test
Here is the test to reproduce 
```rust
#[test]
fn resource_index_load_store() {
    let temp_dir = TempDir::new("arklib_test")
        .expect("Failed to create temporary directory");
    let temp_dir = temp_dir.into_path();

    create_file_at(
        temp_dir.to_owned(),
        Some(FILE_SIZE_1),
        Some(FILE_NAME_1),
    );
    let index = ResourceIndex::build(temp_dir.to_owned());


    index
        .store()
        .expect("Should store index successfully");


    let loaded_index = ResourceIndex::load(temp_dir.to_owned())
        .expect("Should load index successfully");

    // Assert that the loaded index is equal to the original index
    assert_eq!(index, loaded_index);
}
```

When I run this test, I get this output
```bash
thread 'index::tests::resource_index_load_store' panicked at src/index.rs:824:9:
assertion `left == right` failed
  left: ResourceIndex { id2path: {ResourceId { data_size: 10, crc32: 3817498742 }: "/private/var/folders/fg/kw558kp56h574pchd9zylc500000gn/T/arklib_test.VcRhfirveNvX/test1.txt"}, path2id: {"/private/var/folders/fg/kw558kp56h574pchd9zylc500000gn/T/arklib_test.VcRhfirveNvX/test1.txt": IndexEntry { modified: SystemTime { tv_sec: 1709340915, tv_nsec: 542881544 }, id: ResourceId { data_size: 10, crc32: 3817498742 } }}, collisions: {}, root: "/private/var/folders/fg/kw558kp56h574pchd9zylc500000gn/T/arklib_test.VcRhfirveNvX" }
 right: ResourceIndex { id2path: {ResourceId { data_size: 10, crc32: 3817498742 }: "/private/var/folders/fg/kw558kp56h574pchd9zylc500000gn/T/arklib_test.VcRhfirveNvX/test1.txt"}, path2id: {"/private/var/folders/fg/kw558kp56h574pchd9zylc500000gn/T/arklib_test.VcRhfirveNvX/test1.txt": IndexEntry { modified: SystemTime { tv_sec: 1709340915, tv_nsec: 542000000 }, id: ResourceId { data_size: 10, crc32: 3817498742 } }}, collisions: {}, root: "/private/var/folders/fg/kw558kp56h574pchd9zylc500000gn/T/arklib_test.VcRhfirveNvX" }
```

Notice how `tv_nsec` is different

src/index.rs
tareknaser: > ResourceIndex::provide() is the starting point for any user/app. It combines index loading and updating in single step.

Got it
Thanks for clearing this out. 

> Name forget_id is confusing now. 


The initial idea for this PR was to rewrite the file `src/index.rs` as it is and then implement the changes to replace #72 
Do you prefer to include all the changes in the current PR?

src/index.rs
kirillt: Looks like rounding.. If you could investigate why is it rounded, it would be great.

src/index.rs
kirillt: Yeah... It's difficult to keep in mind several PRs because the scope is pretty big.

Let's incorporate useful parts of #72 and unit tests from there into this PR.

src/index.rs
tareknaser: I think I understand now
We're currently storing and loading timestamps in milliseconds, but `SystemTime` relies on `Timespec` under the hood which is:
```rust
struct Timespec {
    tv_sec: i64,
    tv_nsec: Nanoseconds,
}
```
I confirmed locally that if we store and parse the timestamp as nanoseconds, we don't run into this issue. I'll double-check before pushing the code.

src/index.rs
tareknaser: Opened a ticket #91 to further discuss this

src/index.rs
tareknaser: Yes 
Reverted this change

src/index.rs
tareknaser: After spending some time modifying this PR based on #72, I've realized that it might not be a good idea to combine all the changes from both PRs into a single PR. #72 has significant API changes for `ResourceIndex`, and I think we should discuss these changes further in #91 before proceeding.

This PR focuses only on refactoring certain parts of the code to enhance readability. It doesn't alter any logic for `arklib` so it might be a good idea to keep it simple.
What do you think?

src/index.rs
tareknaser: The changes in forget_id in #72 don't work on their own because `ResourceIndex::forget_entry()` and `IndexUpdate` were also modified.

src/index.rs
tareknaser: > I would return just `deleted` collection, because `added` is always empty here.
  
That's why I suggest finalizing the API schema in #91 before implementing changes in a separate PR. #72 modifies `IndexUpdate`, adds its `Default` implementation, and integrates it into multiple API methods. 

Trying to merge #72 changes into this PR while making further modifications along the way could become complicated.

kirillt:
About change from milliseconds to nanoseconds in timestamps: we should double-check that we really can have nanoseconds on mobile platforms. I don't remember precisely, but it seems we had an issue with nanoseconds due to implicit rounding to milliseconds. Could be so even on Linux.

We should investigate why this threshold was introduced: https://github.com/ARK-Builders/arklib/commit/267b844f3a0f04586c7d6178180607fc3cb0fa0d#diff-701633ed548410624cfcae90d38a986adbd60deb5aeedd92355fe2e7702a9c18L34

src/index.rs
```diff
@@ -609,7 +609,7 @@ impl ResourceIndex {
                 "Any collision must involve at least 2 resources"
             );
             *collisions -= 1;
-            if *collisions <= 1 {
+            if *collisions == 1 {
                 self.collisions.remove(&old_id);
```
kirillt, line 613: Should we panic on negative values then?

src/storage/meta.rs
```diff
@@ -38,6 +38,7 @@ pub fn store_metadata<
 }
 
 /// The file must exist if this method is called
+#[allow(dead_code)]
 pub fn load_raw_metadata<P: AsRef<Path>>(
```
kirillt, line 42: This function can be removed together with its unit tests.

src/index.rs
tareknaser: Yes. We should never reach this case

tareknaser: It's of type `usize` so it would never be negative eitherway

tareknaser:
> About change from milliseconds to nanoseconds in timestamps: we should double-check that we really can have nanoseconds on mobile platforms. I don't remember precisely, but it seems we had an issue with nanoseconds due to implicit rounding to milliseconds. Could be so even on Linux.


Before making any changes to `arklib` on how time is stored, we need to understand the impact of storing time in nanosecond on mobile systems. 



If we must keep the millisecond precision, we have two options to consider:
- We can keep the current implementation, which stores time in milliseconds. Only problem is that `IndexEntry` objects won't be equal when writing to ark file and then reading from it.
- We can track only milliseconds when retrieving `last_modified` of each entry. This can be done by converting the `SystemTime` to milliseconds in `scan_entry` as follows:
```rust
let duration = modified
        .duration_since(UNIX_EPOCH)
        .expect("SystemTime before UNIX EPOCH!")
        .as_millis();
let modified =
	UNIX_EPOCH + std::time::Duration::from_millis(duration as u64);
```

kirillt:
@tareknaser we could use nanoseconds, but:
1. Do we really need such a precision? Can several filesystem writes happen with time difference of e.g. 10ns? I hope milliseconds is enough.
2. We also should care about compatibility between different platforms. I'm not sure that Android provides developers with timestamps better than milliseconds.

I could imagine us needing microseconds to timestamp realistically filesystem events. Would we have the rounding issues in this case?

> We can track only milliseconds when retrieving `last_modified` of each entry. This can be done by converting the `SystemTime` to milliseconds in `scan_entry`

This sounds like the way to go.

> We can keep the current implementation, which stores time in milliseconds. Only problem is that IndexEntry objects won't be equal when writing to ark file and then reading from it.

Is this sill a problem if we go the way with converting to milliseconds in `scan_entry`?

tareknaser:
> Do we really need such a precision? Can several filesystem writes happen with time difference of e.g. 10ns? I hope milliseconds is enough.

Yes I think milliseconds is enough. Only issue is https://github.com/ARK-Builders/arklib/pull/87#discussion_r1509775947 

> Is this sill a problem if we go the way with converting to milliseconds in scan_entry?

No. that should solve it

kirillt (approved):

PR #90, opened 2024-02-26
Generic Resource Identifiers with `ResourceIdTrait`

### LINKED ISSUE #75: Index generic over hash function
We can abstract `ResourceIndex` over the id type and, consequently, over hashing function.
```
#[derive(Eq, Ord, PartialEq, PartialOrd, Hash, Clone, Debug)]
pub struct IndexEntry<Id> {
    pub modified: SystemTime,
    pub id: Id,
}

#[derive(PartialEq, Clone, Debug)]
pub struct ResourceIndex<Id> {
    pub id2path: HashMap<Id, CanonicalPathBuf>,
    pub path2id: HashMap<CanonicalPathBuf, IndexEntry<Id>>,

    pub collisions: HashMap<Id, usize>,
    root: PathBuf,
}
```

This way, we can use cryptographic hash functions to test index in simpler cases, when no collisions are present. This can simplify development and debugging. We could also use fake hash functions for testing only collisions.

Cryptographic hash function can also be used for apps where safety is more important than performance. The fast hash functions will be used in experimental "fast mode", it's the most useful for file browser apps.

tareknaser:
## Description
This PR aims to make `ResourceIndex` type **generic** over the **hash function** used to generate resource identifiers

Resolves #75
## Changes
- A new trait is implemented named `ResourceIdTrait`. This trait defines a set of functions and requirements that all specific `ResourceId` implementations must implement
- Within `ResourceIdTrait`, a type alias named `HashType` is defined. This allows different `ResourceId` implementations to use different types for their hash values, but we ensure that all `HashType` types implement specific traits for consistent behavior
- Modify `ResourceIdCrc32` type to implement the `ResourceIdTrait`
## Future work
- `src/index.rs` file heavily relies on the assumption of only using `ResourceIdCrc32`. It will require adjustments, potentially including abstracting some tests, to function with various `ResourceId` types.
- We can add conditional compilation easily. for example we can use `ResourceIdBlake3` that implements `ResourceIdTrait` 
- Later, when https://github.com/ARK-Builders/ark-rust/pull/5 is merged, we can move `src/resource` directory into a separate crate for better organization.

### DIFF
--- a/src/resource/crc32.rs
+++ b/src/resource/crc32.rs
@@ -57,10 +57,6 @@ impl FromStr for ResourceId {
 impl ResourceIdTrait<'_> for ResourceId {
     type HashType = u32;
 
-    fn get_hash(&self) -> Self::HashType {
-        self.hash
-    }
-
     fn compute<P: AsRef<Path>>(data_size: u64, file_path: P) -> Result<Self> {
         log::trace!(
             "[compute] file {} with size {} mb",
@@ -140,12 +136,12 @@ mod tests {
             .len();
 
         let id1 = ResourceId::compute(data_size, file_path).unwrap();
-        assert_eq!(id1.get_hash(), 0x342a3d4a);
+        assert_eq!(id1.hash, 0x342a3d4a);
         assert_eq!(id1.data_size, 128760);
 
         let raw_bytes = fs::read(file_path).unwrap();
         let id2 = ResourceId::compute_bytes(raw_bytes.as_slice()).unwrap();
-        assert_eq!(id2.get_hash(), 0x342a3d4a);
+        assert_eq!(id2.hash, 0x342a3d4a);
         assert_eq!(id2.data_size, 128760);
     }
 
--- a/src/resource/crc32.rs
+++ b/src/resource/crc32.rs
@@ -16,6 +16,9 @@ const KILOBYTE: u64 = 1024;
 const MEGABYTE: u64 = 1024 * KILOBYTE;
 const BUFFER_CAPACITY: usize = 512 * KILOBYTE as usize;
 
+/// Represents a resource identifier using the CRC32 algorithm.
+///
+/// Uses `crc32fast` crate to compute the hash value.
 #[derive(
     Eq,
     Ord,
--- a/src/resource/crc32.rs
+++ b/src/resource/crc32.rs
@@ -30,12 +30,12 @@ const BUFFER_CAPACITY: usize = 512 * KILOBYTE as usize;
 )]
 pub struct ResourceIdCrc32 {
     pub data_size: u64,
-    pub crc32: u32,
+    pub hash: u32,
 }
 
 impl Display for ResourceIdCrc32 {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        write!(f, "{}-{}", self.data_size, self.crc32)
+        write!(f, "{}-{}", self.data_size, self.hash)
     }
 }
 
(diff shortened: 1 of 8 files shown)


tareknaser:
Take a look at https://github.com/ARK-Builders/arklib/tree/blake3_ResourceIdTrait. It shows how easy it is to include a new `ResourceId` type that implements `ResourceIdTrait`, featuring Blake3 as the hash function in this instance.
Also notice that `ResourceIndex` still has collision tracking although blake3 is a cryptographic hash function and there should be no collisions.

src/resource/mod.rs
```diff
@@ -0,0 +1,64 @@
...
+
+pub use crc32::ResourceIdCrc32 as ResourceId;
+
+/// This trait defines a generic type representing a resource identifier.
+///
+/// Resources are identified by a hash value, which is computed from the resource's data.
+/// The hash value is used to uniquely identify the resource.
+///
+/// Implementors of this trait must provide a way to compute the hash value from the resource's data.
+pub trait ResourceIdTrait<'de>:
```
kirillt, line 22: What do you think if we just use `ResourceId`? It seems to be more readable and easier to type when there are many id parameters in the code.

src/resource/crc32.rs
```diff
@@ -23,34 +31,37 @@ use crate::{ArklibError, Result};
     Deserialize,
     Serialize,
 )]
-pub struct ResourceId {
+pub struct ResourceIdCrc32 {
```
kirillt, line 34: Probably the types implementing the trait could be named `ResourceId`, too?

In the usage spot we always can use `crc32::ResourceId`. It seems to be more readable to me.

src/resource/crc32.rs
```diff
@@ -23,34 +31,37 @@ use crate::{ArklibError, Result};
...
-    pub fn compute<P: AsRef<Path>>(
-        data_size: u64,
-        file_path: P,
-    ) -> Result<Self> {
+impl ResourceIdTrait<'_> for ResourceIdCrc32 {
+    type HashType = u32;
+
+    fn get_hash(&self) -> Self::HashType {
+        self.hash
+    }
```
kirillt, line 62: Not sure that we need to expose this. Probably ids should be "black boxes", i.e. users should only use them (via `Eq`, `PartialEq`, `Ord`, `Display` traits) without going into the details of how the ids are constructed.

tareknaser: Currently, `ResourceId` has `data_size` and `hash` as public fields and I see them used in many places all over the codebase (especially in unit tests). 
I thought if the hash is so frequently used, might be a good idea to add a method to get it. 

What do you think?

src/resource/mod.rs
tareknaser: If we rename `ResourceIdCrc32` to `ResourceId`, we can't rename `ResourceIdTrait` to `ResourceId` as well. 

But this shouldn't be an issue since `ResourceIdTrait` is mainly used to impose restrictions on `ResourceId` types so I don't anticipate frequent interactions with it.

src/resource/crc32.rs
kirillt: Let's consider this by use-cases:
1. Unit tests should be able to access internal fields, but they should not be exposed to users/apps.
2. Users/apps could use `data_size` but if they need it, they would need it with other hash functions, too. It's better to provide such a field in some other way, generically for any hash function. Since we just pass this data to crc32 generation function, we could  save it somewhere around. I don't think that users really need to do something with `hash` value, only with the id as a whole.

src/resource/mod.rs
kirillt: Got it. OK, let's see. Probably we'll do one more renaming later.

src/resource/crc32.rs
tareknaser: > Unit tests should be able to access internal fields

In the current setup, implementers of `ResourceIdTrait` have the freedom to decide whether to expose `data_size` and `hash` or not. It may be best not to enforce methods like `get_hash` in the trait itself.

Same goes for any methods related to `data_size`, considering that this field might not be applicable in all implementations of `ResourceIdTrait`, especially in cryptographic hash function implementations.

tareknaser: We didn't include `data_size` in the draft for `BLAKE3` implementation. Here https://github.com/ARK-Builders/arklib/pull/73#discussion_r1468890690

kirillt (approved):
Looks good

